Issue #808, opened 2014-10-21
Inconsistent Download Location with FlexGet


sasese:
Given [this config.yml](http://pastebin.com/3Af4kQZr), pyLoad will alternately download files to their specified folders (for example: X:\Intended\Folder) or to a number of different folders in pyLoads default download folder (for example D:\default pyload download folder\Package Name, or D:\default pyload download folder\XIntendedFolder). When the package is inspected in pyLoad's webui, it shows the correct folder, and if the packages is restarted, it will download to the correct folder.
The occurrence of this issue is unpredictable: the same config.yml run back to back several times results in different outcomes.

Unfortunately neither FlexGet's trace log nor pyLoad's debug logs indicate the or any related problem, but as pyLoad does record the intended download folder correctly it would seem that the error is in pyLoad.

Semantically the folder location seems correct, as far as FlexGet goes. A number of variations have been tried, including forward slashing the path, without avail.

The most curious part of the entire problem is its intermittent nature: sometimes the download folder is correct, sometimes it is not. Sometimes, correct and incorrect folders appear in packages added in the same task, and even in the same package. Any of the three download locations defined in the config.yml, linked to above, may produce the error.

pyLoad 0.49
Windows (7, 32bit, Ultimate)

snickers2k:
Hey. Oh boy, it's been a long time. I absolutely had no free time in the past months, so i'm really sorry.
But i'm still struggling with this issue.

Has your fix already been merged ? And if the crypter.py is still the problem, and you haven't any experience with it, maybe @vuolter can take a look?

thanks guys

vuolter:
My 0.4.10 is currently broken: first I need to fix it, update plugins to lastest in stable, then I can examine that pr. It's a long story...

snickers2k:
i really don't want to bother.. but maybe any news on this?^^

PR #866, opened 2014-11-09
Bug Fixes for Inconsistent Download Location with FlexGet

sasese:
The commits in this pull request resolve both bugs which cause [Inconsistent Download Locations with FlexGet](https://github.com/pyload/pyload/issues/808) (issue #808).

The modification of pyPackage resolves the first bug, which causes download locations passed by FlexGet to be mangled from (for example) X:\Intended\Folder to D:\Default pyLoad Download Folder\XIntendedFolder.

The modifications of the API and FlexGet's pyload output plugin resolve the second bug, which causes download locations passed by FlexGet to be mangled from (for example) X:\Intended\Folder to D:\Default pyLoad Download Folder\Package Name. The modifications of the API are such that they maintain compatibility with FlexGet for installations which have not been updated to the new pyload output plugin, although the fix won't work until it is.

### DIFF
--- a/module/PyPackage.py
+++ b/module/PyPackage.py
@@ -39,8 +39,8 @@ class PyPackage():
 
     @property
     def folder(self):
-        return self._folder # Modification returns self._folder without stripping characters via save_path
-        
+        return self._folder
+
     def toDict(self):
         """ Returns a dictionary representation of the data.
 


snickers2k:
any updates on this? i would even pay to have a working workaround before 0.4.11 ^^ waiting already over a year :(

Roburetto:
Two years down the road these two bug fixes are still need and prove to work with most recent Flexget pyload plugin,

To cut a long story short, one should replace the files from the 2011 release with these two:

[pyload/module/Api.py ](https://github.com/sasese/pyload/blob/a108d9e824e4a0f5d5b1672d642d2f793bebbf0b/module/Api.py) - a835336  on Nov 15, 2014

[pyload/module/PyPackage.py](https://github.com/sasese/pyload/blob/ef704fe52bb9eb58ae3c615f50c1fce03827982b/module/PyPackage.py) - ef704fe  on Nov 16, 2014

timofonic:
@Roburetto Even in the Git code?

I see there's *TONS* of unresolved pull requests :(

Roburetto:
@timofonic I have since found that fix a835336 actually BREAKS URL RECOGNITION with Flexget:
Flexget: "No suited urls in entry"
So you better stick with original Api.py.

Roburetto:
@timofonic It would seem so. If you compare with  Apy.pi from stable branch you will see the fixes are not applied there.

PR #889, opened 2014-11-15
Fix/web module URIs through reverse proxy

robbi:
Hi,
This patch fixes accessing to the web interface through a reverse proxy and might fix some prefix bug. I've tested it using nginx frontend.
Here is the nginx configuration extract I use:

```
    location /pyload {
        proxy_set_header Host $host;
        proxy_pass https://localhost:8000/pyload;
        proxy_redirect default;
    }
```

And the pyload.conf extract:

```
webinterface - "Webinterface":
    str template : "Template" = default
    bool activated : "Activated" = True
    str prefix : "Path Prefix" = /pyload
    builtin;threaded;fastcgi;lightweight server : "Server" = threaded
    ip host : "IP" = 127.0.0.1
    bool https : "Use HTTPS" = True
    int port : "Port" = 8000
```

Regards

### DIFF
--- a/module/HookManager.py
+++ b/module/HookManager.py
@@ -141,8 +141,8 @@ class HookManager:
                 if self.core.debug:
                     traceback.print_exc()
 
-        self.log.info(_("Activated plugins: %s") % ", ".join(sorted(active)))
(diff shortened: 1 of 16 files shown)


I-RiX-I:
https://github.com/pyload/pyload/pull/2015

Issue #976, opened 2014-12-25
FilecryptCc.py "no captcha found"


Gutz-Pilz:
25.12.2014 10:49:21 INFO      Decrypting starts: http://filecrypt.cc/Container/CD088C4FE9.html
25.12.2014 10:49:21 DEBUG     FilecryptCc: No captcha found

zapp-brannigan:
@vuolter 
Do you have any clue how to solve this captchas?
https://cloud.githubusercontent.com/assets/10416328/5628496/3dc53b48-95a8-11e4-91e4-49eca4192f4e.PNG

Jdownloader2 is able to handle it. I've looked into JD-code but i don't understand it :S

Ghoststage:
Ok THX :((

2015-01-06 14:27 GMT+01:00 zapp-brannigan <notifications@github.com>:

Gutz-Pilz:
http://stackoverflow.com/questions/8380972/recaptcha-in-django-without-forms
http://stackoverflow.com/questions/1440239/how-to-use-python-plugin-recaptcha-client-for-validation
^^maybe this helps??

Ghoststage:
Unfortunately, I do not understand :(

Leider verstehe ich das nicht :(

zapp-brannigan:
Some "good" news:
Today i've inspected JDownloader2 code again and looked at the http requests which are made by JD. I've borrowed some code for our plugin, but unfortunately it still doesn't work.
But as of today JD2's support for recaptchaV2 is also broken ;)

vuolter:
50% done, see recaptcha in captchaservice. ;)

zapp-brannigan:
@vuolter 
I have a working recaptchaV2 solution. How is the progress on your side? Shall i open a PR for Captchaservice.py or are you almost done?

vuolter:
Mi side is out-of-order! XD
Your PR will be welcome. ;)

Ghoststage:
VERY BIG THX

PR #1043, opened 2015-01-14
[CaptchaService.py] Added support for ReCaptchaV2

zapp-brannigan:
Fix for https://github.com/pyload/pyload/issues/976


zapp-brannigan:
Something is messed up here :( I've taken the latest CaptchaService.py from my local installation as base.
I've only added the RecaptchaV2-class and the 2 import statements...

PR #1006, opened 2015-01-03
Redesigned webinterface with bootstrap css

mariusbaumann:
I now found time to redesign the webinterface of pyload. Probably you like it. 

General changes:
- New style on every page
- New head-menu
- Completely new icon-set
- New status lables with color for eatch status
- New styled progress bars
- Completely redesigned and reformated settings page
- Redesigned overlay-windows
- Some html formate changes 
- More responsive design (not on queue page, there i sitll have some problems with the javastcript generated ui)

Important: I have nothing changed on the functionality of pyload, only the frontend of the webinterface has changed. On some pages i had to make little changes on the javascript to generate the html in the right structure (on exsample for the new icon i had to change the tags from img to span so i had to change the javascript to reacting on the span tags) 

I hope you like it :) 

Now some screenshots:

bildschirmfoto 2015-01-03 um 07 07 55

bildschirmfoto 2015-01-03 um 07 08 11

bildschirmfoto 2015-01-03 um 07 08 52

bildschirmfoto 2015-01-03 um 07 09 10

bildschirmfoto 2015-01-03 um 07 08 29

PR #1071, opened 2015-01-23
Fix support for the 'Path Prefix' option on the default template

gimco:
I`ve replaced all the absolute uri on the default template to use the 'url' filter to take care of the "Path Prefix" option of the webinterface.


I-RiX-I:
Will this require rework of the new bootstrap webinterface?

gimco:
If you're referring to the pull #1006, yes, It will require rework on the new webinterface.
If the legacy pyload webinteface will be deprecated in favor of the new bootstrap, you should reject this pull and I'll send a new pull with the modifications on the (new official) bootstrap theme.

I-RiX-I:
Yes indeed, seems like that needs a fix too, its paths are hardcoded.

I-RiX-I:
I reworked some things: (unfinished)

https://github.com/pyload/pyload/pull/2015

GammaC0de:
https://github.com/pyload/pyload/pull/3077

Issue #1096, opened 2015-01-29
uploaded.to: problems with checkErrors() for premium users, no downloads with premium account


oliverrahner:
Downloading with version 0.78 of the UploadedTo hoster plugin throws "File not downloadable for free users" error.
I haven't fully understood how the plugin handles premium downloads, but all cases seem to go through checkErrors(). "Current waiting period" is only there for free users, so the plugin exits with self.fail()

oliverrahner:
There are several checks in checkErrors() that are only valid for free users, maybe this method should be split.

oliverrahner:
Did a quick and dirty fix for the downloading issue in 7a8b6edcfba91e2ecf5652c4f8bde9c6ec925fc8

Killer1337:
After update to 0.79 I still get the same error

oliverrahner:
"The same error" being "File not downloadable for free users"?

Killer1337:
yes.

oliverrahner:
Sorry, but I couldn't find any mention whether you have an uploaded.to premium account?

Killer1337:
No I don't

oliverrahner:
Yes, because my problem was that I do have a premium account, and it gave me the same error message, which makes even less sense than in your case.
I'm only a "normal user" and thought I'd make the fix that helped me available to others :)
But I will take a quick look whether I can replicate your issue.

PR #1103, opened 2015-01-30
hoster uploaded.to: only run checkErrors() for free accounts

oliverrahner:
Solves the issue discussed in #1096, and potentially other issues.


Killer1337:
I don't speak python very well so this might be a dumb question, but pleas enlighten me.
How does this fix the issue for non-premium users?

oliverrahner:
It doesn't.
I only learned that free users also have problems after I made this pull request.

oliverrahner:
Won't work, because the parent class SimpleHoster already calls checkErrors(), so that filtering inside that method can't be worked around.

PR #1114, opened 2015-01-30
after finalizing the package

sixteeN11:
there is the possibility the movie mover run after the package has been completed is ? because currently it starts with me only when you start the pyLoad . but it runs day and night with me by.. thx


vuolter:
Is it a question?

immenz:
I think he is asking for the feature that the hook starts after the packagefinished event. 

May I suggest that I post my customized EpisodeMover hook in the near future? I made some major changes and @Rusk85 stopped developing on it, so I kept my changes to myself. But I think this Plugin needs some updates.

vuolter:
@immenz Open a new PR when your ready (no hurry). ;)

PR #1124, opened 2015-02-01
Captcha9Kw should be initialized activated false

immenz:
Captcha9Kw should be initialized activated false.


vuolter:
https://github.com/pyload/pyload/commit/60394c74c92775303094a1069917d53169ab2fcd

PR #1130, opened 2015-02-02
Fix XdadevelopersCom.py

zapp-brannigan:
I know the "self.download" line looks strange but the "get = {}" parameter is not working on this website.


stickell:
If I remember well it has already been fixed in 0.4.10

vuolter:
Done: https://github.com/pyload/pyload/commit/e275aeb2c829f34b57670f55f05cc1a1b4e7f259
I completely forgot about this core issue. Now I have bad feelings about other plugins... :fearful:

PR #1141, opened 2015-02-03
Addressed some Extractor Issues

immenz:
Some Changes that fix little Problems.

Some other still remain:
- after Extraction the remainig parts are logged as skipped -> ugly
- sevenzip is absolutely damaged, cant get it working
- can't find a list of supported File formats for Unrar, so that this can be defined precisely 
- also it seems not to extract when package finished (although option Wait for all downloads to be finished: is false)

### DIFF
--- a/module/plugins/hooks/ExtractArchive.py
+++ b/module/plugins/hooks/ExtractArchive.py
@@ -63,7 +63,10 @@ class ArchiveQueue(object):
 
 
     def get(self):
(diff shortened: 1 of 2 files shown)


immenz:
Also the interval for executing periodically (I think the only purpose for this is to react 'instant' on package finished) should be set lower and/or at least make configurable vie settings

PR #1156, opened 2015-02-07
Avoid an expected error

ItachiSan:
Running the FastCGI *flup* server launch a ***ValueError*** as the webserver thread does not run as the main one and it can't catch keyboard shut-down commands (ex. CTRL+C).
The webserver is stopped properly (should be) by the PyLoad main thread, so the exception is expected.


vuolter:
https://github.com/vuolter/pyload/commit/dc65f738913827ddf38a91f046b92e519ebd677a

Issue #1191, opened 2015-02-19
ZippyShare keeps changing its checksum


sebdelsol:
Based on what @gaberad suggested, I have Zippyshare working for at least the 4 last changes. 

Body's JS scripts are evaluated with all the DOM statements turned into regular variables initialized with their fetched values in the html.
```python
# get all the scripts inside the html body
tree = lxml.html.fromstring(self.html)
scripts = tree.xpath('//body//script[@type="text/javascript"]/text()')

# meant to be populated with the initialization of all the DOM elements found in the scripts
initScripts = []

def repl(element):
    id, attr = element.group(1), element.group(4)      # attr might be None

    varName = '%s_%s' %(id, attr)

    initValues = tree.xpath('//*[@id="%s"]/@%s' %(id, attr))
    initValue = '"%s"' %initValues[-1] if initValues else 'null'

    initScripts.append('var %s = %s;' %(varName, initValue))
    return varName

# handle all getElementById
reVar = r'document.getElementById\([\'"](\w+)[\'"]\)(\.)?(getAttribute\([\'"])?(\w+)?([\'"]\))?'
scripts = [re.sub(reVar, repl, script) for script in scripts]

# get the file's url by evaluating all the scripts
return  JSE_chainEval(initScripts + scripts + ['dlbutton_href'])
```

Some JS scripts are meant to fail at some point, hence  `JSE_chainEval()` needs to catch errors and resume the evaluation of all the remaining scripts with *the same JS context*:

```python
def JSE_chainEval(scripts):
    rt = PyV8.JSContext()
    rt.enter()

    for script in scripts:
        try:
            ret = rt.eval(script.encode('utf8') if isinstance(script, unicode) else script)
        except:
            pass

    return ret
```
EDIT : I don't know how to use xpath (or beautifulSoup) in a hoster. So I just assumed that lxml.html would be available in some way...

EDIT2: ZippyShare changed again! I had to handle some DOM introspection, eg. document.getElementById('omg').getAttribute('class')

gaberad:
Thanks for the solution @sebdelsol.

FYI:
I don't have PyV8 installed and just wanted to get something working for now, so I changed JSE_chainEval to use self.js.eval. It works for me, but obviously doesn't have the error catching that the original does. Ended with this (relevant parts):
```
...
import lxml.html

class ZippyshareCom(SimpleHoster):
...
    def handleFree(self, pyfile):
        recaptcha   = ReCaptcha(self)
[...]

PR #1202, opened 2015-02-23
Zippyshare correction (bug #1191)

sebdelsol:
This version doesn't handle deliberate errors in the JS script(s). 
I don't want to add a dependency on PyV8 that is not needed _at the moment_.


sebdelsol:
I've found a way to handle deliberate errors in the JS scripts without relying on PyV8. So I'll close this pr.

PR #1205, opened 2015-02-23
Zippyshare bug #1191 correction

sebdelsol:
This version use the internal `Beatifulsoup`. When you'll decide to get rid of it, I suggest to add a dependency on `lxml`.

I've found a way to handle deliberate errors in the JS scripts without relying on `PyV8` by adding JS `try/catch` statements around all the zippyshare scripts.


vuolter:
Thanks! :smile:

sebdelsol:
You're welcome. I'll keep an eye on zippyshare, I guess they've got some other tricks to pull on...

PR #1218, opened 2015-03-02
fix traffic left

OndrejIT:


stickell:
Have you seen `parseFileSize` in `module.utils`?

OndrejIT:
yes, but i don't see Terabyte...

stickell:
You are right. I upgraded it to support even Exabyte but will be in 0.4.10

OndrejIT:
ok...

PR #1227, opened 2015-03-06
[ExternalScripts] Encode unicode characters

jellysheep:
Hi,
I don't know if this is the right place to submit patches to pyload, but I would like to submit a small fix:
If a file with unicode characters in the filename is downloaded ExternalScripts.py fails with e.g. this error:

    ExternalScripts: Error in download_finished.sh: 'ascii' codec can't decode byte 0xc3 in position 29: ordinal not in range(128)

This commit makes the ExternalScripts plugin encode unicode characters to UTF-8.


vuolter:
THX.

Issue #1246, opened 2015-03-10
Zippyshare has changed again


sebdelsol:
broken again...

PR #1247, opened 2015-03-10
[zippyshare] bug correction #1246

sebdelsol:
Handle funky introspection like :
```javascript
test = {
    test: function(x, y) {
        document.getElementById(x).href = y;
   }

test.test('dlbutton', "http:\\ZippyshareIsFullOfTrickster.com");
```


sraedler:
I have speedissues with Zippyshare since this update.
I just downloader about 5 kb...

can you proof this ?

sebdelsol:
This version of the plugin has just changed how the file url is extracted from the website. 
Since there's only one file url to be read on the website, the change in this code could not make any difference in the way the file is downloaded afterward... 
Have tried to compare how it downloads with your browser ?

sraedler:
Ok, now it is fast again, sorry....

I tried in Browser and it was fast like always...

The problem was just after I restarted and pulled, so I thought it is maybe because of the Plugin.

Issue #984, opened 2014-12-28
[PremiumizeMe] Sometimes not working


Gutz-Pilz:
28.12.2014 11:43:57 ERROR     PremiumizeMe: 
28.12.2014 11:43:57 ERROR     PremiumizeMe: No hoster loaded

guessing that the issue is since:
https://github.com/pyload/pyload/commit/b72be64cc16660e3ba877d97dbc0e717bb030666#diff-5cb8b7e5d9fcdfa73d07da5793a9e2e3

tseroo:
At least it is partly working now, but the main issue is, that lots of hosters are missing! As you can see from the logs, e.g. uploaded.to is not available. Direct download of uploaded.to links via web browser are working with full speed though!

606	01.01.2015 12:08:36	DEBUG	PremiumizeMe: Regexp: .*(asfile\.com|depfile\.com|mixturecloud\.com|movshare\.net|purevid\.com|shareflare\.net|uploadto\.us|videoweed\.es|vip-file\.com).*
605	01.01.2015 12:08:36	DEBUG	PremiumizeMe: New hosters: asfile.com, depfile.com, mixturecloud.com, movshare.net, purevid.com, shareflare.net, uploadto.us, videoweed.es, vip-file.com
604	01.01.2015 12:08:36	DEBUG	PremiumizeMe: Overwritten hosters: BitshareCom, DepositfilesCom, FilepostCom, FilerNet, FilerioCom, FourSharedCom, LetitbitNet, NetloadIn, NowVideoSx, OneFichierCom, RapidgatorNet, StreamcloudEu, TurbobitNet, UploadableCh, UploadedTo, UptoboxCom

vuolter:
604 01.01.2015 12:08:36 DEBUG PremiumizeMe: Overwritten hosters: BitshareCom, DepositfilesCom, FilepostCom, FilerNet, FilerioCom, FourSharedCom, LetitbitNet, NetloadIn, NowVideoSx, OneFichierCom, RapidgatorNet, StreamcloudEu, TurbobitNet, UploadableCh, __UploadedTo__, UptoboxCom

kasperj:
Everything is working. At least for me it does ;)

vuolter:
So all working... can I smoothly close this issue report now?

tseroo:
@vuolter 

Of course you're right ... overwritten means it was already there before!! 
Problem was that it did not use PremiumizeMe for downloading. Only after disabling/enabling it, it was working again as expected.

[...]

PR #1250, opened 2015-03-12
Fix filename for OboomCom downloads in conjunction with PremiumizeMe

r15ch13:
Should fix the OboomCom filename problem, mentioned by vikwin in #984


vuolter:
I'm working to fix the name detection issue in the core class. ;)

r15ch13:
Sounds great! I leave this open, as a temporary workaround. :)

vuolter:
Please try if works now.

Issue #1230, opened 2015-03-07
[ExtractArchive] Recycle instead of Delete


vuolter:
Forum: http://forum.pyload.org/viewtopic.php?f=13&t=3943

immenz:
Maybe that's helpful: https://github.com/hsoft/send2trash

vicegold:
This broke extracting archives on Mac OSX. send2trash is installed but not working. Becuase of that months old bug of "pluginsettings not saved" bug in pyLoad the option get's activated after every download again. So i can't turn it off. 
```22166	13.04.2015 12:40:32	ERROR	ExtractArchive: test.part18.rar | Unknown error | global name 'send2trash' is not defined
22167	13.04.2015 12:40:32	ERROR	ExtractArchive: test.part18.rar | Extract failed```

vuolter:
@vicegold Please open as new issue.

PR #1285, opened 2015-03-25
[ExtractArchive] Send2Trash Integration

immenz:
https://github.com/pyload/pyload/issues/1230

Integrated https://github.com/hsoft/send2trash

With this you don't have to make s2t a dependency for pyload. Maybe you want to do that, but this is just a quick'n'dirty solution.

install using pip: 

    pip install send2trash

Tested on windows. Did not effect systems where send2trash isn't installed, except an Errormessage in Log if it isn't installed but selected in config

### DIFF
--- a/module/plugins/hooks/ExtractArchive.py
+++ b/module/plugins/hooks/ExtractArchive.py
@@ -106,24 +106,24 @@ class ArchiveQueue(object):
 class ExtractArchive(Hook):
     __name__    = "ExtractArchive"
     __type__    = "hook"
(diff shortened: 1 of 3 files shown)


I-RiX-I:
Thanks for integrating this.

How would I include this feature in my Windows setup? I am using the officially compiled version of pyLoad, that runs without python itself installed.

vuolter:
Install pip: http://stackoverflow.com/questions/4750806/how-to-install-pip-on-windows
Type: `pip install Send2Trash`

Issue #1311, opened 2015-04-03
Zippyshare download file.html


Mister-XY:
Hello, i found a small bug if any download via zippyshare. In some downloads, you get only a file.html file instead of the realy filename.
Here is an example to test
3616	03.04.2015 23:13:26	DEBUG	ReCaptcha: Key not found
3615	03.04.2015 23:13:26	DEBUG	ZippyshareCom: Handled as free download
3614	03.04.2015 23:13:26	DEBUG	ZippyshareCom: File status: online
3613	03.04.2015 23:13:26	DEBUG	ZippyshareCom: File name: tvp-starwarsrebels-s01e01e02-720p.part1.rar | File size: 155189248 byte
3612	03.04.2015 23:13:26	DEBUG	ZippyshareCom: Current file info: {'status': 2, 'name': u'tvp-starwarsrebels-s01e01e02-720p.part1.rar', 'url': u'http://www56.zippyshare.com/v/JzrRg8mW/file.html', 'pattern': {'S': u'148', 'U': u'MB', 'KEY': u'JzrRg8mW', 'N': u'tvp-starwarsrebels-s01e01e02-720p.part1.rar'}, 'size': 155189248.0}
3611	03.04.2015 23:13:26	DEBUG	ZippyshareCom: Previous file info: {'status': 2, 'name': u'tvp-starwarsrebels-s01e01e02-720p.part1.rar', 'url': u'http://www56.zippyshare.com/v/JzrRg8mW/file.html', 'pattern': {'S': u'148', 'U': u'MB', 'KEY': u'JzrRg8mW', 'N': u'tvp-starwarsrebels-s01e01e02-720p.part1.rar'}, 'size': 155189248.0}
3610	03.04.2015 23:13:26	DEBUG	ZippyshareCom: Update file info...
3609	03.04.2015 23:13:26	DEBUG	ZippyshareCom: File name: tvp-starwarsrebels-s01e01e02-720p.part1.rar | File size: 155189248 byte
3608	03.04.2015 23:13:26	DEBUG	ZippyshareCom: Current file info: {'status': 2, 'name': u'tvp-starwarsrebels-s01e01e02-720p.part1.rar', 'url': u'http://www56.zippyshare.com/v/JzrRg8mW/file.html', 'pattern': {'S': u'148', 'U': u'MB', 'KEY': u'JzrRg8mW', 'N': u'tvp-starwarsrebels-s01e01e02-720p.part1.rar'}, 'size': 155189248.0}
[...]

PR #1314, opened 2015-04-04
[zippyshare] fix bug #1311

sebdelsol:
Zippyshare sometimes encode the file name with an image.
In this case, use the file link to get the file name.
(it's probably a bad hack. I don't know nothing about pyload Core).

### DIFF
--- a/module/plugins/hoster/ZippyshareCom.py
+++ b/module/plugins/hoster/ZippyshareCom.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import re
+from urllib import unquote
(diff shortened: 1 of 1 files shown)


sebdelsol:
Since there are new cases where the file name could only be found in the javascript, I believe it's way more robust (and simpler) to extract the filename from the actual file url than trying to extract it from the javascript with a regular expression.

Mister-XY:
with this plugin, i can download via zippyshare with the correct filename insteat of file.html. In the package is show me also as file.html, but when the download start, i have the correct filename.

PR #1335, opened 2015-04-09
Fix network signature and Content-Disposition

### LINKED ISSUE #1334: HTTP Signature
Hi all,
With commit 55073d5073b6cf80bdb865f79546e871d9bbcbc2 which try to solve file name issues, the solution introduce a signature on HTTP Request. Requests are done with User-Agent

```
Python-urllib/2.7
```

Even the hook UserAgentSwitcher will not work because the request done on [BasePlugin.py](https://github.com/pyload/pyload/blob/stable/module/plugins/hoster/BasePlugin.py#L45) bypass request done by pycurl.

I'm doing a fix right now which manage Content-Disposition parsing in [HTTPChunk.py](https://github.com/pyload/pyload/blob/stable/module/network/HTTPChunk.py#L252).

If you could give me some example of Content-Disposition header to manage, it will help me.

MeegoPy:
Fix #1334

### DIFF
--- a/module/network/HTTPChunk.py
+++ b/module/network/HTTPChunk.py
@@ -19,7 +19,7 @@
 from os import remove, stat, fsync
 from os.path import exists
 from time import sleep
-from re import search
+import re
 from module.utils import fs_encode
 import codecs
 import pycurl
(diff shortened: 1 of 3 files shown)


vuolter:
Stable core is freezed. Please, check if 0.4.10 is affected, then reopen it here: https://github.com/vuolter/pyload/pulls

Issue #1343, opened 2015-04-10
lolabits.es


nickiman:
Hi,
I would like request your support to add to pyLoad this new hoster:

http://lolabits.es/

It´s a new hoster that has no captchas and also has a very good bandwidth to download files .

I attach you several download links that I get:

http://lolabits.es/juanvilma/3lH0bb1t2014juanvilma.part006(1),22368568.rar(archive)
http://lolabits.es/juanvilma/3lH0bb1t2014juanvilma.part007(1),22368651.rar(archive)
http://lolabits.es/juanvilma/3lH0bb1t2014juanvilma.part008(1),22368630.rar(archive)
http://lolabits.es/juanvilma/3lH0bb1t2014juanvilma.part009(1),22368702.rar(archive)
http://lolabits.es/juanvilma/3lH0bb1t2014juanvilma.part01(1),22368641.rev
http://lolabits.es/juanvilma/3lH0bb1t2014juanvilma.part010(1),22368647.rar(archive)

I attach you also the output of the log in pyLoad just in case can be useful for you:
```
3	10.04.2015 21:06:03	DEBUG	Redirected import module.plugins.hoster.BasePlugin -> userplugins.hoster.BasePlugin
4	10.04.2015 21:06:03	INFO	Download starts: http://lolabits.es/juanvilma/3lH0bb1t2014juanvilma.part006(1),22368568.rar(archive)
5	10.04.2015 21:06:03	DEBUG	UserAgentSwitcher: Use custom user-agent string: Mozilla/5.0 (Windows NT 6.1; WOW64; rv:37.0) Gecko/20100101 Firefox/37.0
6	10.04.2015 21:06:03	DEBUG	BasePlugin: Redirect #0 to: http://lolabits.es/juanvilma/3lH0bb1t2014juanvilma.part006(1),22368568.rar(archive)
7	10.04.2015 21:06:04	WARNING	Download failed: 3lH0bb1t2014juanvilma.part006(1),22368568.rar(archive) | File not found
8	10.04.2015 21:06:04	INFO	UnSkipOnFail: Looking for skipped duplicates of: 3lH0bb1t2014juanvilma.part006(1),22368568.rar(archive) (pid:34)
9	10.04.2015 21:06:04	INFO	UnSkipOnFail: No duplicates found
10	10.04.2015 21:06:05	INFO	Download starts: http://lolabits.es/juanvilma/3lH0bb1t2014juanvilma.part007(1),22368651.rar(archive)
[...]

PR #1345, opened 2015-04-11
[New Hoster] lolabits.es

zapp-brannigan:
As requested in https://github.com/pyload/pyload/issues/1343


vuolter:
Thank you for your support always. :D

PR #1370, opened 2015-04-24
[UpleaCom] Fix folder handling

GammaC0de:


GammaC0de:
Hi,
Please notice that the changes to `SimpleCrypter.py` are not mandatory, if you don't want to make changes to `SimpleCrypter.py`, you can update UpleaCom.py only and it would work OK.


The changes to `SimpleCrypter.py` were to address `XFSCrypter.py` code `LINK_PATTERN = r'<(?:td|TD).*?>\s*<a href="(.+?)".*?>.+?(?:</a>)?\s*</(?:td|TD)>'` when the html TD would span multiple lines, eg.:
```html
<td class="title-link">
<img style="margin-top:-3px;" src="/img/file.png"/>
<a href="http://uplea.com/dl/A3B488FFCF93F3F">http://uplea.com/dl/A3B488FFCF93F3F</a>
</td>
``` 
Without using `re.S` in `SimpleCrypter.py`'s `getLinks`, LINK_PATTERN would not match this link..
Alternative solution would be to override `getLinks` inside `XFSCrypter.py` using `re.S` so making changes in `XFSCrypter.py` only and not `SimpleCrypter.py`.

GammaC0de:
new pr: #1375

PR #1379, opened 2015-04-26
[ExtractArchive] [AntiVirus] fix send2trash

### LINKED ISSUE #1351: send2trash not working on OSX
The integration of send2trash broke extracting archives on Mac OSX. 

send2trash is installed but not working. Because of that months old bug of "pluginsettings not saved" bug in pyLoad the option get's activated after every download again. So i can't turn it off. (Maybe it should be turned off by default?)

```22166 13.04.2015 12:40:32 ERROR ExtractArchive: test.part18.rar | Unknown error | global name 'send2trash' is not defined 
22167 13.04.2015 12:40:32 ERROR ExtractArchive: test.part18.rar | Extract failed```

The archives will be extracted, but not deleted and the hooks that would be called after extraction won't work because the extraction fails.

GammaC0de:
fix #1351

### DIFF
--- a/module/plugins/hooks/AntiVirus.py
+++ b/module/plugins/hooks/AntiVirus.py
@@ -11,7 +11,7 @@ from module.utils import fs_encode, save_join
 class AntiVirus(Hook):
     __name__    = "AntiVirus"
     __type__    = "hook"
-    __version__ = "0.07"
+    __version__ = "0.08"
 
     #@TODO: add trash option (use Send2Trash lib)
     __config__ = [("action"    , "Antivirus default;Delete;Quarantine", "Manage infected files"                     , "Antivirus default"),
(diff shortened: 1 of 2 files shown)


vuolter:
@GammaC0de `import send2trash` is called on `setup`, why calling it again should solve #1351?

GammaC0de:
Hi,
Since `import send2trash` is called on `setup` and not globally as all other imports, it is relevant only for `setup`
Calling `send2trash`  at `scan` or `_extract` cases an unknown function exception (I've verified that)

Since you cannot import send2trash globally, the solution is to import it on `scan` and `_extract` after you know `trashable` is True so its safe to import `send2trash`

vuolter:
You're right: https://github.com/pyload/pyload/commit/711621f3d3a59f7ec5c2684dfac921e2c5902563
Thanks!

GammaC0de:
:+1:

PR #1385, opened 2015-04-29
[UploadedTo] detect maintenance

### LINKED ISSUE #1383: Feature Request: [UploadedNet] Detect Maintenance Mode
Please update the UploadedNet plugin so it detects uploaded maintenance mode.

Adding a timer to re-check indefinitely every 5 minutes or so might be necessary.

Else all links in the queue are skipped as offline and have to be restarted manually.

Here is the html code if Uploaded is in maintenance

http://paste2.org/OaBy4vZ6

EDiT: A check for the head title "uploaded.net - Maintenance - Wartungsarbeiten" should suffice, I guess..

GammaC0de:
fix #1383


I-RiX-I:
Thank you :+1:

PR #1390, opened 2015-05-03
suggestion [XFSCrypter] use HTMLParser instead of regular expressions

GammaC0de:
Hi,
I would like to suggest using HTMLParser  for grabbing links inside html <TD> tags
This way it is more robust  and chances for false positive are near zero.
even after last commit, XFSCrypter does not catch cases like:
```
<td class="title-link">
<a href="http://uplea.com/dl/A3B488FFCF93F3F">http://uplea.com/dl/A3B488FFCF93F3F</a>
<img style="margin-top:-3px;" src="/img/file.png"/>
</td>
```
Performance is not degraded. 
So using  HTMLParser is better in my opinion.

What do you say?

### DIFF
--- a/module/plugins/internal/XFSCrypter.py
+++ b/module/plugins/internal/XFSCrypter.py
@@ -53,9 +53,6 @@ class XFSCrypter(SimpleCrypter):
 
     URL_REPLACEMENTS = [(r'&?per_page=\d+', ""), (r'[?/&]+$', ""), (r'(.+/[^?]+)$', r'\1?'), (r'$', r'&per_page=10000')]
 
-    LINK_PATTERN = r'<a href="(.+?)".*?>.+?(?:</a>)?\s*</(?:td|TD)>'
-    NAME_PATTERN = r'<[tT]itle>.*?\: (?P<N>.+) folder</[tT]itle>'
-
     OFFLINE_PATTERN      = r'>\s*\w+ (Not Found|file (was|has been) removed)'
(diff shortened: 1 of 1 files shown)


vuolter:
I want to explain why this PR was rejected (for now):
XFSCrypter is a wide used class, like SimpleHoster. For this kind of plugins we can not simply remove a key attribute like LINK_PATTERN, because many other plugins use it, and would go down otherwise. In future, if LINK_PATTERN becomes too complex, we may combine current regex behaviour with HTMLParser. ;)

GammaC0de:
Ok, thats fine, this is why I wrote `suggestion` in the title :wink: 
one connent though:
the new LINK_PATTERN you added a new group but I think should use an unnamed group, ie:
`<a href="(.+?)".*?>.+?(?:</a>)?\s*(?:<.+>\s*)?</(?:td|TD)>`
```patch
-<a href="(.+?)".*?>.+?(?:</a>)?\s*(<.+>\s*)?</(?:td|TD)>
+<a href="(.+?)".*?>.+?(?:</a>)?\s*(?:<.+>\s*)?</(?:td|TD)>
```
or even better use  `<(?:td|TD).*?>\s*(?:<.+>\s*)?<a href="(.+?)".*?>.+?(?:</a>)?\s*(?:<.+>\s*)?</(?:td|TD)` which is better in my opinion.

vuolter:
https://github.com/pyload/pyload/commit/7b095378cba22aa2cfadfd62699d77bf1724da41

GammaC0de:
:thumbsup:

PR #1405, opened 2015-05-09
[ClickAndLoad] fix bug when webinterface uses SSL

### LINKED ISSUE #1387: Click And Load not working when "Use HTTPS" for the Webinterface is activated
This is a plugin issue with Click and Load:

When the option "Use HTTPS" for the Webinterface is activated (-> webserver: threaded for SSL support instead of builtin), Click and Load won't work.
Unfortunately, this setting also affects the Click and Load function, although Click and Load listens on a different Port (9666) than the webinterface (8000).

When the option is activated, Click and Load only accepts HTTPS requests. But almost all Click and Load requests connect with http://-ip-:9666/flash/, and not with https://-ip-:9666/flash/

The result is a "400 Bad Request" response from the Click and Load function to the client, Example:
Remote Address:-ip-:9666, Request URL:http://-ip-:9666/flash/addcrypted2
Status Code:400 Bad Request
Response: "The client sent a plain HTTP request, but this server only speaks HTTPS on this port."

-> "Use HTTPS" for the Webinterface should not affect Click and Load.
Click and Load should always listen to plain http requests on port 9666, regardless of whether the webinterface uses https or not.

Turning off "Use HTTPS" for the Webinterface and everything works fine, but that's not really a solution regarding security concerns.

Reproduce:
install pyload on debian (tested on Raspbian), activate SSL /  "Use HTTPS" Webinterface and try to add a link with click and load. (I can add an exact list of steps to reproduce if necessary)

GammaC0de:
fix #1387

The solution requires python's SSL lib which  is officially supported in python 2.6+  but there is an installable version for 2.5 (https://pypi.python.org/pypi/ssl/)

### DIFF
--- a/module/plugins/hooks/ClickAndLoad.py
+++ b/module/plugins/hooks/ClickAndLoad.py
@@ -3,6 +3,11 @@
 import socket
 import time
 
+try:
+    import ssl
+except ImportError:
+    pass
+
 from threading import Lock
 
 from module.plugins.Hook import Hook, threaded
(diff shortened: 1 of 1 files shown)


vuolter:
:+1:

PR #1426, opened 2015-05-25
[NitroFlare] Re-Added RECAPTCHA_KEY

immenz:
Key is necessary, because cannot be obtained automatically (also the variable is referred some lines further in the code)


immenz:
Also updated Wait pattern & removed Error pattern, because the wait message contains the error message, and so when a wait time is given download always failed.

PR #1438, opened 2015-05-30
Remove obsolete AdYouLike call

lazlev:
@vuolter @tmerle 

AdYouLike is currently not used anymore on the hoster's site. Additionally the regex had to be edited to actually grab the domain value. With these changes I could download again until yesterday's update of the plugin.

Seems like this commit isn't actually working: https://github.com/pyload/pyload/commit/6616c00ba1c40f5d56959bd0e4725f26250e1292#diff-5d57a6bc31e3d4fa2d3010afe37c2d46

The original sample link still works for testing:
http://dl.free.fr/b2ANA3gPp


lazlev:
Should I add another pull request for the missing self.download(...) line or will you fix it yourself?

vuolter:
A new PR is welcome.

PR #1474, opened 2015-06-09
Ordered configuration files

Nippey:
Reflect order of config files in (web-) interfaces to allow further structuring in future.


Nippey:
My commit 20facaf4 shows the way to use of ordered dictionaries.
You may or may not use it. :)

I'm always open to feedback!

RaNaN:
Something very similar is already implemented in 1.0.0 branch, we can't merge it into stable unfortunatly.

PR #1482, opened 2015-06-10
fix #1475

### LINKED ISSUE #1475: External script package finished
Hello, is it possible that only the in external script the package_finished folder work with script?
I have also copy the script into the Folder package_extracted and to test in archive_extrackted and pyload load on start the script but if an download is extracted, the scripts did not run. Only the Script in Package_finished did run.

GammaC0de:


GammaC0de:
Apparently, this was caused by 9f3ab5, other modules are surely effected.
Hook manager checks for  `event_list` declaration on the constructor so setting it on `setup` is too late and would not set a hook.

PR #1486, opened 2015-06-11
Update SimpleHoster.py

GammaC0de:
Allows plugins to use code like:
```python
try:
    pass

except Exception, e:
    self.error(e)
```


GammaC0de:
https://github.com/pyload/pyload/commit/ea27af7361bb75d665c29120d9cdc254a209afa8

PR #1521, opened 2015-06-17
Plugins for the Webinterface

Nippey:
Hi,

this pull request is the preparation for a bigger plan of mine and should be a discussion platform, of how my ideas are implemented best!

My plan is to have the possibility to extend the Webinterface with additional content.

I made up a function name `webinterface_add_plugin_content`, that --if exposed-- will be recognised by `pyload_app.py`.

This function can return a list of tuples that will be added to the Webinterface as HTML tags.

    {"type":"javascript", "content": "alert('Hello!');"},

will add the following right before the HEAD tag:

    <script type="text/javascript">alert('Hello!');</script>

1. Is the way I added the functionality acceptable? (Specially named function)
2. Do you accept, at all, that plugins may alter the Web-IF?

These possibilities may tempt others to link to 3rd party websites to include scripts into the Web-IF, therefore I would like to add the possibility for Plugins to serve whole files for the Web-IF _locally_.
Currently there is only the `/api/call` which can only serve JSON compatibe responses.
For Images or Javascript, this won't work.
I all cases I want to prevent serving additional files together with the Plugins, so everything to be served is embedded into the `.py` file.

I considered one of the following possibilities:

A. 
  - The plugin itself creates the files in `<configDir>/tmp/plugins/<plugin>`. 
  - The plugin is responsible to keep the folder clean.
  - The Webinterface gets a new route to serve files from this folder.

B.
  - The plugins provides a variable named `__web_files__` which is parsed by the `PluginManager`
  - `PluginManager` cleans the folder on startup and refills it with every plugin that provides the variable.
  - The PluginManager uses REGEX to parse these variables, so not every character allowed. BASE64-Encode?
  - The Webinterface gets a new route to serve files from this folder.

C.
  - Add a new API function that specifically routes to a special Plugin-Function to return RAW text `/api/pluginContent/<pluginName>/<fileName>`

I would prefer either B or C.
What do you guys think?

_PS: I accidentally added a commit (08fb26f) which already has a pull request to this branch. I will remove it next time I have shell access._


RaNaN:
Thank you for the pull request. First there is to say we can not accept such patches on the stable branch, because this one is only for plugin fixes.

The webUI in stable can't be considered very good and extensible, we had plans to make a new one from scratch, which can be found in the 1.0.0 branch. I would prefer to extend this one and finish it instead of patching and adding stuff to the old one.

Allowing plugins to extend the webUI is certainly interesting, but could you describe a few use-cases for what you think such plugins could be useful?

nomad71:
Actually I have at least one:

forum.pyload.org/viewtopic.php?f=13&t=4061&p=17244

But I could also imagine having it display the remaining drive space or quotas for premium accounts.
You could also have (custom) buttons to quickly change settings like the speed limit or toggle send2trash. Manually triggering a plugin-update would be awesome, too.

And - not to be rude - but the web interfaces ability to manage the packages is fairly primitive and even a pain in the ass when it comes to managing more complex tasks like removing the links of a specific hoster from a specific folder - there could be a plugin for those more specialized tasks as well.

Although I'm fairly certain, that improving on use cases like the latter is part of the reason you think of rewriting the interface from scratch, consider this:

pyLoad is highly modular anyway - why not making the web interface modular as well? People will find a way to make use of it, that's for sure!

Nippey:
One plugin that I'm using for some months now are HTML5-Notifications for the Webinterface.

notify

I started changing `pyLoadCore.py` and many more files at first until I realised that (almost) everything was already available on a Plugin-Level.

I finished the implementation of Web-IF plugins locally.
As soon as it is polished, I will add my code to this pull request, maybe there are some interesting things hidden.

And another thing I was thinking about for a longer time now is downloading from Serienjunkies.org
Serienjunkies has a lot of parameters and I wish to have some interactivity to select the Season/Episode/Quality/Hoster while creating a package.

Nippey:
Finally added the code.
Code says more than words!

Feel free to browse and of course integrate it! :)

vuolter:
Very nice work! :+1:

Gutz-Pilz:
>And another thing I was thinking about for a longer time now is downloading from Serienjunkies.org
Serienjunkies has a lot of parameters and I wish to have some interactivity to select the Season/Episode/Quality/Hoster while creating a package.

^i'd really appreciate this!
https://github.com/Gutz-Pilz/pyLoad-stuff

PR #1534, opened 2015-06-23
Update/Fix GamefrontCom hoster plugin

frederikmoellers:
GameFront uses a `<span>` tag around long filenames. This caused the
plugin to include the tag in the filename.
Instead of the (possibly abbreviated) filename in the page's content,
the title (HTML `<title>` tag) is now used.
Use this link to see the issue (file is not really a RAR archive, just a file with random content): http://www.gamefront.com/files/25080153/thisisalongfilenameitssolongicantevenseetheendwowanditsstillgoingonahthereistheend.rar


vuolter:
:+1:

PR #1615, opened 2015-07-30
[FilecryptCc.py] Fix https://github.com/pyload/pyload/issues/1614

zapp-brannigan:
Fix https://github.com/pyload/pyload/issues/1614


vuolter:
:+1:

PR #1666, opened 2015-08-02
[ZDF] Encode unicode characters when loading XML

jellysheep:
The XML file from the ZDF website often contains unicode characters, which lead to errors like
`'ascii' codec can't decode byte 0xc3 in position 29: ordinal not in range(128)`. This change makes the plugin encode unicode characters to UTF-8 before parsing the XML data.

### DIFF
--- a/module/plugins/hoster/ZDF.py
+++ b/module/plugins/hoster/ZDF.py
@@ -42,7 +42,7 @@ class ZDF(Hoster):
 
 
     def process(self, pyfile):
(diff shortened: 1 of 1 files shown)


vuolter:
Did you try `xml = etree.fromstring(self.load(self.XML_API % self.get_id(pyfile.url), decode=False))`?

jellysheep:
Sorry for the delay, I didn't have the time to test it.
Your solution is working as well and seems to be a cleaner way of doing it; I didn't know this was possible. Should I file another pull request?

jellysheep:
Ok, thanks for 2cf928d.

PR #1701, opened 2015-08-04
Fix XFSHoster.py - set_cookie should be imported

mobo95:


vuolter:
https://github.com/pyload/pyload/commit/f80576bd8c809932dd9f62483402ec18532699b2

PR #1708, opened 2015-08-04
Download deleted because of file size mismatch

estaban:
One some hosters like OneFichier for example, the webpage displays a file size of 1.46GB.
However the real file size can be 1.4649853GB, which makes a difference between the downloaded file size and the expected file size.

size_tolerance is meant for that, but its value is currently set to 1MB. This is an issue in the above case where the difference between expected size and downloaded size is almost 10MB. In many cases, this causes the downloaded file to simply be deleted, even though it has been downloaded without any issue.

The biggest possible size difference is 10MB, at least if the website displays 2 decimals and that the biggest unit is GB. Therefore I propose to change the max tolerance from 1MB to 10MB.

FYI the error message in the log looks like that:
`Fail: File size mismatch | Expected file size: 1492501135.36 | Downloaded file size: 1488025852`

### DIFF
--- a/module/plugins/internal/SimpleHoster.py
+++ b/module/plugins/internal/SimpleHoster.py
@@ -310,8 +310,8 @@ class SimpleHoster(Hoster):
             self.captcha.invalid()
             self.retry(10, reason=_("Wrong captcha"))
 
-        # 10485760 is 10MB, tolerance is used when comparing displayed size to real size
-        # For example displayed size can be 1.46GB for example, but real size can be 1.4649853
+        # 10485760 is 10MB, tolerance is used when comparing displayed size on the hoster website to real size
+        # For example displayed size can be 1.46GB for example, but real size can be 1.4649853GB
         elif self.check_download({'Empty file': re.compile(r'\A((.|)(\2|\s)*)\Z')},
                                  file_size=self.info['size'] if 'size' in self.info else 0,
                                  size_tolerance=10485760,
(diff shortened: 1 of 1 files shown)


LePastis:
Hey , I think he misses incrementing the version number for the file to be taken by the updater

bast69:
I have the same error for all downloads with Uptobox

LePastis:
Sorry, i don't see commit 4323507752c54c03cf109ee4c5b74d11bae753dd
So, it's work fine for me. @mobo95 @bast69 Check if you have 1.80 version of SimpleHoster.py in your userplugin folder

mobo95:
I've 1.80 version, now the file has been downloaded without errors

janpal:
Does not work for me. Seems that the difference in size is ~35 MB.

janpal:
I am on 1.80 and i am still getting the error:
05.08.2015 21:49:15 WARNING Download failed: Rozpravkova jazda.part1 29 casti sk.rar | File size mismatch | Expected file size: 1800000000.0 | Downloaded file size: 1835008000

Seems that diff is too big - 35 MB which goes over the treshold in the code.

Perhaps a stupid recommendation... What about skipping the size check if the size of downloaded file is bigger than the expected size?

original download link: http://ulozto.sk/xF8ry6W4/rozpravkova-jazda-part1-29-casti-sk-rar

bast69:
ok, all works fine. tks guys

Issue #1635, opened 2015-07-31
ExternalScripts


DKeppi:
External scripts are not being called since last 2 updates!

Normally at the DEBUG or LOG there is an Info where a Script was found, mine is the folder all_archives_extracted, but since last update this info is gone and no script is called

Web-Eye:
same here

nomad71:
Ay! Can confirm!

Gummibaer:
The External scripts dont see the Scriptfiles !! Nothing in Log

CrEaK:
Can confirm! External scripts dont see my script at startup. No entry in the logs

c0rrre:
Can confirm too

dizzle667:
can confirm it. I have it with package_finished. If it helps, AndroidPhoneNotify sends a message, that one package could not be completed, although it is completed successfully. Maybe there is something wrong with the completion?

PR #1712, opened 2015-08-05
[ExternalScripts] Prepend directory to filenames

jellysheep:
As `os.path.join()` returns the basename of the files, the directory has to be prepended.


jellysheep:
This should fix #1635.

PR #1738, opened 2015-08-09
FIX for 'ReCaptcha' object has no attribute 'COOKIES'

Velociraptor85:
FIX for known Problem, Version change for UpdateManager


Velociraptor85:
since internal/Plugin v0.30 fixed

PR #1747, opened 2015-08-12
Fixed Rapidgator account plugin.

fayf:
Small fix, presumably for an API endpoint URL change on Rapidgator's side.


fayf:
Commit string doesn't hook into Github issues. Will resubmit.

PR #1784, opened 2015-08-25
Fix Click'N'Load bug

frederikmoellers:
Using unencrypted C'N'L would give an HTTP 500 error because the
function add() was called with no parameter.


frederikmoellers:
For testing purposes, you can try the examples on
http://jdownloader.org/knowledge/wiki/glossary/cnl2
Plain CNL2 is on the bottom

Alexlambertz:
@vuolter What didn't you like at this implementation? The number of merge commits? I guess the fix is needed and fits into the general code in this module.

pascalschmiederer:
Jdownloader Click n Load on JD Homepage doesnt work too.

Velociraptor85:
works for me

GammaC0de:
@Velociraptor85 Although is works, `add` function should not take any parameter so this PR is still needed.

Velociraptor85:
@GammaC0de yes I tested it without the test parameter as the PR suggests so this PR does work according to my tests

milaq:
@frederikmoellers are you still onto this?

frederikmoellers:
Somewhat... I have stopped using pyLoad some time ago so I have no reason to continue working on it. The PR is imho still good and I'm alright with merging it.
If the CLA check is necessary, then I'm afraid I won't bother to accept it (I can elaborate on the reasons, but this is not the place to do so). However, anyone can just copy my changes and create a new PR.

lukastribus:
@GammaC0de @Velociraptor85 I'm not sure I understand, everyone seems to agree that this change fixes the problem, what is the reason this cannot be committed?

This is a trivial one-line fix, a CLA is hardly necessary.

PR #1785, opened 2015-08-25
[Plugin.py] cannot log unicode chars

GammaC0de:
if you try, for example, `self.log_debug(u"\u2022")` you get:
```
  File "/usr/lib/python2.7/logging/__init__.py", line 874, in emit
    stream.write(fs % msg.encode("UTF-8"))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 53: ordinal not in range(128)
Logged from file Plugin.py, line 188
```

### DIFF
--- a/module/plugins/internal/Plugin.py
+++ b/module/plugins/internal/Plugin.py
@@ -146,7 +146,7 @@ def chunks(iterable, size):
 class Plugin(object):
     __name__    = "Plugin"
     __type__    = "hoster"
-    __version__ = "0.30"
+    __version__ = "0.31"
     __status__  = "testing"
 
     __pattern__ = r'^unmatchable$'
@@ -180,7 +180,7 @@ class Plugin(object):
 
     def _log(self, level, plugintype, pluginname, messages):
         log = getattr(self.pyload.log, level)
-        msg = encode(" | ".join((a if isinstance(a, basestring) else str(a)).strip() for a in messages if a))
+        msg = " | ".join((a if isinstance(a, basestring) else str(a)).strip() for a in messages if a)
         log("%(plugintype)s %(pluginname)s%(id)s: %(msg)s"
             % {'plugintype': plugintype.upper(),
                'pluginname': pluginname,


vuolter:
Please retry now.

GammaC0de:
still not working :( 
Just add `self.log_debug(u"\u2022")` to any plugin and see what happens.
```
21.09.2015 09:34:27 DEBUG     HOSTER MegaCoNz[94]: Decrypted Attr: MEGA{"n":"Test;File1.txt","c":"BoczK9XIuGRi0w0ivEWcFwTkNftV"}
  File "/usr/lib/python2.7/logging/__init__.py", line 874, in emit
    stream.write(fs % msg.encode("UTF-8"))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 51: ordinal not in range(128)
Logged from file Plugin.py, line 218
  File "/usr/lib/python2.7/logging/__init__.py", line 874, in emit
    stream.write(fs % msg.encode("UTF-8"))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 51: ordinal not in range(128)
Logged from file Plugin.py, line 218
```

vuolter:
Because I decide to delay the update on last minute. So no commit was pushed. XD

GammaC0de:
:+1:

GammaC0de:
Unfortunately I get the same result even with the recent commits.
this code works OK though (`decode` instead of `encode`)
```python
msg = u" | ".join(decode(a).strip() for a in messages if a)
```

GammaC0de:
I see it was fixed in `Hoster.py` but not in `Plugin.py` [Plugin.py#L214](https://github.com/pyload/pyload/blob/stable/module/plugins/internal/Plugin.py#L214)
Also, Shouldn't `Hoster` call its base class (`Plugin`) `_log()` function?

vuolter:
Good eye!
Hoster _log is lightly different than Plugin one. ;)

Issue #1749, opened 2015-08-12
[Hoster] LolabitsEs - 'str' object is not callable


nickiman:
Hi,
I´m having this issue with the Lolabits.es hoster...
![image](https://cloud.githubusercontent.com/assets/3627902/9222464/4264c3b2-40f2-11e5-8a71-28a3a0f5c26d.png)

I attach you the output of pyLoad in Debug mode:
```
12.08.2015 12:39:56	INFO	Download starts: LaLigaJusticiajuanvilma.part04.rar
12.08.2015 12:39:56	DEBUG	HOOK UserAgentSwitcher: Use custom user-agent string: Mozilla/5.0 (Windows NT 6.1; WOW64; rv:38.0) Gecko/20100101 Firefox/38.0
12.08.2015 12:39:56	DEBUG	HOOK UserAgentSwitcher: Use custom user-agent string: Mozilla/5.0 (Windows NT 6.1; WOW64; rv:38.0) Gecko/20100101 Firefox/38.0
12.08.2015 12:39:56	DEBUG	HOSTER LolabitsEs[785]: PROCESS URL http://lolabits.es/juanvilma2/LaLigaJusticiajuanvilma.part04,38656156.rar(archive) | PLUGIN VERSION 0.03
12.08.2015 12:39:56	DEBUG	HOSTER LolabitsEs[785]: Config option `use_premium` not found, use default `True`
12.08.2015 12:39:56	INFO	HOSTER LolabitsEs[785]: Updating file info...
12.08.2015 12:39:56	DEBUG	HOSTER LolabitsEs[785]: File info: {'status': 2, 'url': u'http://lolabits.es/juanvilma2/LaLigaJusticiajuanvilma.part04,38656156.rar(archive)', 'size': 136314880.0, 'name': u'LaLigaJusticiajuanvilma.part04.rar', 'pattern': {'S': u'130', 'U': u'MB', 'N': u'LaLigaJusticiajuanvilma.part04.rar'}}
12.08.2015 12:39:56	DEBUG	HOSTER LolabitsEs[785]: Previous file info: {}
12.08.2015 12:39:56	INFO	HOSTER LolabitsEs[785]: File name: LaLigaJusticiajuanvilma.part04.rar
12.08.2015 12:39:56	INFO	HOSTER LolabitsEs[785]: File size: 136314880 bytes
12.08.2015 12:39:56	INFO	HOSTER LolabitsEs[785]: File status: online
[...]

PR #1788, opened 2015-08-26
[Plugin.py] fix #1787 and #1749

### LINKED ISSUE #1787: [bug] cannot use self.load(url, decode=...)
Hi,
There is a bug which prevents using `load()` with the `decode="encoding"` parameter:
if you specify `self.load(pyfile.url, decode="unicode-escape")` , for example, you get this error:
```
  File "/usr/share/pyload/module/PluginThread.py", line 191, in run
    pyfile.plugin.preprocessing(self)
  File "/home/user/.pyload-0.4.9/userplugins/internal/Hoster.py", line 185, in preprocessing
    return self.process(self.pyfile)
  File "/home/user/.pyload-0.4.9/userplugins/internal/SimpleHoster.py", line 288, in process
    self.handle_free(pyfile)
  File "/home/user/.pyload-0.4.9/userplugins/hoster/LolabitsEs.py", line 45, in handle_free
    decode="unicode-escape")
  File "/home/user/.pyload-0.4.9/userplugins/internal/Plugin.py", line 354, in load
    res = decode(res, decode)
TypeError: 'str' object is not callable
```
This is because of [Plugin.py#L348](https://github.com/pyload/pyload/blob/stable/module/plugins/internal/Plugin.py#L348):
 `res = decode(res, decode)`

Since the parameter `decode` contains a `str` it has overwritten the global function `decode` an so the call at line 348 fails because `decode` is a `str` and not a function.

GammaC0de:
fix #1787 and #1749

### DIFF
--- a/module/plugins/internal/Plugin.py
+++ b/module/plugins/internal/Plugin.py
@@ -7,6 +7,7 @@ import inspect
 import os
 import re
 import urllib
+import sys
 
 if os.name != "nt":
     import grp
@@ -146,7 +147,7 @@ def chunks(iterable, size):
 class Plugin(object):
     __name__    = "Plugin"
     __type__    = "hoster"
-    __version__ = "0.30"
+    __version__ = "0.32"
     __status__  = "testing"
 
     __pattern__ = r'^unmatchable$'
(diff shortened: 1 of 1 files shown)


GammaC0de:
@vuolter can you check this please?

vuolter:
Merging...

PR #1823, opened 2015-09-15
[utils.py] Added Petabytes to the formatSize function

mrhanky17:
pyLoad does not start one FreeBSD when having Terabytes of space, adding Petabytes in the list resolves this.


vuolter:
Already in 0.4.10 .

PR #1827, opened 2015-09-19
[UptoboxCom] account update

benbox69:
The UptoboxCom account plug-in is not working since change has been made in version 0.11...
I tried to make this update conform with the using of XFSAccount... but the login action is completely different because I am using the answer of the ajax request made by Uptobox to login a user and I am not parsing the HTML content to fill a form.
This version is tested and works very well...

### DIFF
--- a/module/plugins/accounts/UptoboxCom.py
+++ b/module/plugins/accounts/UptoboxCom.py
@@ -4,16 +4,16 @@ import json
 import re
 import time
 
-from module.plugins.internal.Account import Account
+from module.plugins.internal.XFSAccount import XFSAccount
 
 
-class UptoboxCom(Account):
+class UptoboxCom(XFSAccount):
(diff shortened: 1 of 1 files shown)


vuolter:
Check https://github.com/pyload/pyload/commit/bd89c11fd385bdbff1f5eb1d7c7d534e6d728946.

benbox69:
Not sure to understand what's wrong with my commit.

vuolter:
The new XFSAccount version parses json object as well, so I just reverted the UptoboxCom as it was before.

benbox69:
Ok I understand but it's not working...
You get a 502 error on line 161 request (XFSAccount) because the login and password are needed... That's why I changed the login action.

vuolter:
@benbox69 Checking...

Issue #1756, opened 2015-08-14
LinkCryptWs - 'bool' object has no attribute 'decrypt'


macross5:
Again an error with an crypter...

> 2458	14.08.2015 23:06:35	INFO	Debug Report written to debug_LinkCryptWs_14-08-2015_23-06-35.zip
2457	14.08.2015 23:06:35	ERROR	Decrypting failed: http://linkcrypt.ws/dir/u9zwn9sbm6nscp1 | 'bool' object has no attribute 'decrypt'
2456	14.08.2015 23:06:35	DEBUG	CRYPTER LinkCryptWs[1570]: Links are captcha protected
2455	14.08.2015 23:06:35	DEBUG	CRYPTER LinkCryptWs[1570]: LOAD URL http://linkcrypt.ws/dir/u9zwn9sbm6nscp1 | cookies=True | get={} | req=None | decode=True | multipart=False | post={} | ref=True | just_header=False
2454	14.08.2015 23:06:35	DEBUG	CRYPTER LinkCryptWs[1570]: PROCESS URL http://linkcrypt.ws/dir/u9zwn9sbm6nscp1 | PLUGIN VERSION 0.10
2453	14.08.2015 23:06:35	DEBUG	HOOK UserAgentSwitcher: Use custom user-agent string: Mozilla/5.0 (Windows NT 6.1; WOW64; rv:38.0) Gecko/20100101 Firefox/38.0
2452	14.08.2015 23:06:35	INFO	Decrypting starts: http://linkcrypt.ws/dir/u9zwn9sbm6nscp1
2451	14.08.2015 23:06:35	DEBUG	Redirected import module.plugins.crypter.LinkCryptWs -> userplugins.crypter.LinkCryptWs

Bambie42:
This https://gist.github.com/Bambie42/989aa620a14d783d2078 fixes the bool problem, but then the plugin always says the captcha is invalid... I think that is a cookie problem?

macross5:
I will try it soon. Thx for your help.

macross5:
you are right. now i cant decrype the captcha as you said.

Bambie42:
Writing a simple decrypter with python-requests solve this issue, so i think this is pyload handling the cookies wrong?

j4m:
I also think it depends on the cookie handling and that this is the reason for #1605 to fail too.

ddbxx:
[...]

PR #1882, opened 2015-09-28
[Captcha] Requse the hoster http request

GammaC0de:
generic fix for #1756:
Requse the hoster http request by default so don't loose the cookies (eg. session cookie)
(Thanks for @ddbxx for spotting this bug)
Useful for `LinkCryptWs` and `FilecryptCC` maybe more


vuolter:
Great PR!

Issue #1865, opened 2015-09-24
HundredEightyUploadCom not downloading file


vilorious:
download creates just 162 B file. Captcha is working, it does show correct filesize and it starts downloading but it doesn´t download.

joberreiter:
Without debug log or error message, we have to switch on our 'mighty crystal ball of fortune-telling' -- but it have to be cleaned prior to use ...
Alright? ;-)

vilorious:
Iam so sorry. I thought your crystal ball is in better shape :P 
Can HundredEightyUploadCom detect Iam using pyload and give me fakefile?

Well I dont find anything in in debug log (but yeah Iam not the best with that stuff). Thank you very much for looking into it!!!

My debug log:

>25.09.2015 16:36:55	INFO	Added package 180upload.com containing 1 links
548	25.09.2015 16:36:55	DEBUG	Fetched 1 values from cache for HundredEightyUploadCom
549	25.09.2015 16:36:55	DEBUG	Finished Info Fetching for HundredEightyUploadCom
550	25.09.2015 16:36:57	INFO	Download starts: How.to.Get.Away.with.Murder.S02E01.720p.HDTV.X264-DIMENSION.mkv
551	25.09.2015 16:36:57	DEBUG	HundredEightyUploadCom: Update file info...
552	25.09.2015 16:36:57	DEBUG	HundredEightyUploadCom: Previous file info: {}
553	25.09.2015 16:36:57	DEBUG	HundredEightyUploadCom: Current file info: {'status': 2, 'url': u'http://180upload.com/7rnhn24voya7', 'size': 810863820.8, 'name': u'How.to.Get.Away.with.Murder.S02E01.720p.HDTV.X264-DIMENSION.mkv', 'pattern': {'S': u'773.3', 'U': u'MB', 'N': u'How.to.Get.Away.with.Murder.S02E01.720p.HDTV.X264-DIMENSION.mkv'}}
554	25.09.2015 16:36:57	DEBUG	HundredEightyUploadCom: File name: How.to.Get.Away.with.Murder.S02E01.720p.HDTV.X264-DIMENSION.mkv | File size: 810863820
555	25.09.2015 16:36:57	DEBUG	HundredEightyUploadCom: File status: online
556	25.09.2015 16:36:58	DEBUG	HundredEightyUploadCom: Update file info...
[...]

Issue #1859, opened 2015-09-24
[uptobox] first argument must be string or compiled pattern


pixou80:
New error  : 
Warning	Download failed : 
"Name of the file" | first argument must be string or compiled pattern

xilmen:
Same error 

WARNING   Download failed : "my file" | first argument must be string or compiled pattern

f6fvy:
This issue now supersedes #1819 !

tribal123:
Same here with Easybytez

jibedo:
Same error

WARNING Download failed : "Name of the file" | first argument must be string or compiled pattern

Bambie42:
Post a debug log and a sample file

jibedo:
My debug log when I begin my download ( link : http://uptobox.com/pbtfx2t7vtqc) : 

>24.09.2015 18:35:19 DEBUG     HOSTER UptoboxCom[52]: File info: {'status': 2, 'url': u'http://uptobox.com/pbtfx2t7vtqc', 'size': 4724464025.6, 'name': u'Air.2015.FRENCH.720p.BluRay.x264-VENUE.Premium.zone-telechargement.com.mkv', 'pattern': {'S': u'4.4', 'U': u'GB', 'N': u'Air.2015.FRENCH.720p.BluRay.x264-VENUE.Premium.zone-telechargement.com.mkv'}}
24.09.2015 18:35:19 DEBUG     HOSTER UptoboxCom[52]: Previous file info: {}
24.09.2015 18:35:19 INFO      HOSTER UptoboxCom[52]: File name: Air.2015.FRENCH.720p.BluRay.x264-VENUE.Premium.zone-telechargement.com.mkv
24.09.2015 18:35:19 INFO      HOSTER UptoboxCom[52]: File size: 4724464025 bytes
24.09.2015 18:35:19 INFO      HOSTER UptoboxCom[52]: File status: online
24.09.2015 18:35:19 DEBUG     HOSTER UptoboxCom[52]: LOAD URL http://uptobox.com/pbtfx2t7vtqc | cookies=True | get={} | req=None | decode=True | multipart=False | post={} | ref=False | just_header=False
24.09.2015 18:35:19 INFO      HOSTER UptoboxCom[52]: Processing as free download...
24.09.2015 18:35:19 DEBUG     HOSTER UptoboxCom[52]: Getting download link #1
[...]

Issue #1852, opened 2015-09-23
"first argument must be string or compiled pattern"


darkpella:
Following to issue #1833 

I just copied all files form the latest authoring from vuolter (commit 9a16506) to my usr/share/pyload folder and started pyLoad.
the " 'xxxxxx' object has no attribute 'user' " issue seems to be gone, but a new issue surfaced, at least with Easybytez.
Now every downloads fails with the following log entries:
"84 23.09.2015 23:31:26 INFO Download starts: XXXX.rar
85 23.09.2015 23:31:26 INFO HOSTER EasybytezCom[1285]: Updating file info...
86 23.09.2015 23:31:29 INFO HOSTER EasybytezCom[1285]: File name: XXXX.rar
87 23.09.2015 23:31:29 INFO HOSTER EasybytezCom[1285]: File size: 525336576 bytes
88 23.09.2015 23:31:29 INFO HOSTER EasybytezCom[1285]: File status: online
89 23.09.2015 23:31:33 INFO HOSTER EasybytezCom[1285]: Processing as free download...
90 23.09.2015 23:31:33 WARNING Download failed: XXXX.rar | first argument must be string or compiled pattern
91 23.09.2015 23:31:33 INFO ADDON UnSkipOnFail: Looking for skipped duplicates of: XXXX.rar (pid:43) "

Please notice that though line 89 states the dwnload will be processed as a free one,I do own a valid premium Easybytez account, I entered its credentials into pyLoad and the "Account" tab under the "Configuration" page (Webinterface) confirms my Easybytez account is a valid premium one.

(in vuolter we trust!)

Exormeter:
Can confirm, premium access seems to be broken in this commit.
Using Share-Onlnie.biz and am not able to use my premium acc.

smystero:
Same here

Hadronius:
Same here with a different host (Worldbytez.com).

Any ideas on how to fix this?!

joberreiter:
@vuolter
Seems like the Hoster (EasyBytez, WorldBytez) has no variable `RECAPTCHA_PATTERN` defined.
In `internal/XFSHoster.py` this variable is initialized with `None`, so the error is a logical result.

Has every hoster a different `RECAPTCHA_PATTERN`?

joberreiter:
Ok, this solves the issue, but now I'm receiving a 'media error' at the webUI-dialog. Hm.

lorenzobb:
Same issue for me.

[...]

PR #1883, opened 2015-09-28
[internal/XFSHoster.py] Add captcha pattern for Recaptcha and Solvemedia

joberreiter:
Fixes the 'first argument must be string or compiled pattern' of #1852, #1859 and #1865.


vuolter:
:+1:

joberreiter:
@vuolter, @GammaC0de:
Is it incorrect to set the captcha patterns for Recaptcha and Solvemedia globally in XFSHoster.py?
Why is #1902 better?

GammaC0de:
although both solutions works, #1902 gets the patterns from `captcha\ReCaptcha.py` and `captcha\SolveMedia.py` which is better because the patterns resides only in one place and if they will change in the future, you only have to change them in one place so there is no chance that you will forget to change them in other places

vuolter:
@joberreiter  Totally agree with @GammaC0de , so I just merged #1902 .

joberreiter:
I see and agree with you -- haven't noticed that those are already set in the captcha plugins.
:thumbsup:

PR #1900, opened 2015-09-30
Strip trailing slashes before case sensitive checking

manuelm:
Strip trailing slashes. Otherwise the last pathname component might be empty


vuolter:
:+1:

PR #1941, opened 2015-10-03
[ClickAndLoad] Update

GammaC0de:
`forward()` function definition error.
also, a `Lock` is not required because function `_server()` is already `threaded`.

P.S.
what does the new setting `"Add packages to queue"` does exactly?
also, why does the delay `time.sleep(10)` before actually calling `_server()`?


GammaC0de:
:+1:

PR #1964, opened 2015-10-06
Double URL Fix

kasperj:
Fix for the issue with http://url.tld/http://url.tld/XXX

### DIFF
--- a/module/plugins/internal/Base.py
+++ b/module/plugins/internal/Base.py
@@ -378,7 +378,7 @@ class Base(Plugin):
 
 
     def fixurl(self, url, baseurl=None, unquote=True):
-        url = fixurl(url, unquote=False)
+        #url = fixurl(url, unquote=False)
 
         if not baseurl:
             baseurl = fixurl(self.pyfile.url)


vuolter:
@kasperj I need some sample url affected by that issue to try to repair the fixurl function. Thanks.
This PR is a temp workaround.

kasperj:
Issue was for nearly every URL, Crypter, Hosters, etc.

http://uploaded.net/file/yhtsb3a6/cpd.1080p.S02E09.part01.rar

joberreiter:
Is it possible to extend the test-cases in /tests (there are already test_api.py and test_json.py).
Or to implement a test-case hook/addon?

vuolter:
@joberreiter In 0.4.9 not, we need a new test suite. Maybe in 0.5.5, because 0.5.3 if full of feature requests! D:

joberreiter:
Regarding url fixing, does this help: https://pymotw.com/2/urlparse/ ?

Notice the end of the page,
```python
from urlparse import urljoin
print urljoin('http://www.example.com/path/file.html', 'anotherfile.html')
print urljoin('http://www.example.com/path/file.html', '../anotherfile.html')
```
which leads to
```
http://www.example.com/path/anotherfile.html
http://www.example.com/anotherfile.html
```

joberreiter:
@vuolter
In `module/plugins/internal/Base.py` there are two functions `fixurl()`, one is defined in the class itself:
```python
    def fixurl(self, url, baseurl=None, unquote=True):
        #url = fixurl(url, unquote=False)

        if not baseurl:
            baseurl = fixurl(self.pyfile.url)

        if not urlparse.urlparse(url).scheme:
            url_p = urlparse.urlparse(baseurl)
            baseurl = "%s://%s" % (url_p.scheme, url_p.netloc)
            url = urlparse.urljoin(baseurl, url)

        return fixurl(url, unquote)
```

And the other one is imported from `internal.Plugins`:
```python
def fixurl(url, unquote=None):
    newurl = urllib.unquote(url)

    if unquote is None:
        unquote = newurl == url

    newurl = html_unescape(decode(newurl).decode('unicode-escape'))
    newurl = re.sub(r'(?<!:)/{2,}', '/', newurl).strip().lstrip('.')

    if not unquote:
        newurl = urllib.quote(newurl)

    return newurl
```
Can this be a possible source of confusion?

vuolter:
@GammaC0de Yep, I need a better name for one of them at least! XD

PR #1980, opened 2015-10-08
fix log_debug() logs even when debug mode off

GammaC0de:


GammaC0de:
never mind, not needed

vuolter:
okay

Issue #1979, opened 2015-10-08
[SimpleHoster|UploadedTo] KeyError: 'ID'


joberreiter:
Seems to be a successor to #1974.

```
08.10.2015 13:42:51 DEBUG     HOSTER UploadedTo[9883]: PROCESS URL http://uploaded.net/file/xxxxxxxx | PLUGIN VERSION 0.97
08.10.2015 13:42:51 INFO      HOSTER UploadedTo[9883]: Updating file info...
08.10.2015 13:42:52 DEBUG     HOSTER UploadedTo[9883]: File info: {'status': 2, 'name': 'KDZEREv.part2.rar', 'url': u'http://uploaded.net/file/xxxxxxxx', 'pattern': {}, 'folder': 'KDZEREv.part2.rar', 'size': 25436706.0}
08.10.2015 13:42:52 DEBUG     HOSTER UploadedTo[9883]: Previous file info: {'status': 2, 'name': 'KDZEREv.part2.rar', 'url': u'http://uploaded.net/file/xxxxxxxx', 'pattern': {}, 'folder': 'KDZEREv.part2.rar', 'size': 25436706.0}
08.10.2015 13:42:52 INFO      HOSTER UploadedTo[9883]: File name: KDZEREv.part2.rar
08.10.2015 13:42:52 INFO      HOSTER UploadedTo[9883]: File size: 25436706 bytes
08.10.2015 13:42:52 INFO      HOSTER UploadedTo[9883]: File status: starting
08.10.2015 13:42:52 DEBUG     HOSTER UploadedTo[9883]: LOAD URL http://uploaded.net/file/xxxxxxxx | redirect=True | cookies=True | get={} | req=None | decode=True | multipart=False | post={} | ref=False | just_header=False
08.10.2015 13:42:52 INFO      HOSTER UploadedTo[9883]: Processing as free download...
08.10.2015 13:42:52 DEBUG     HOSTER UploadedTo[9883]: LOAD URL http://uploaded.net/language/en | redirect=True | cookies=True | get={} | req=None | decode=True | multipart=False | post={} | ref=True | just_header=True
08.10.2015 13:42:52 DEBUG     HOSTER UploadedTo[9883]: LOAD URL http://uploaded.net/js/download.js | redirect=True | cookies=True | get={} | req=None | decode=True | multipart=False | post={} | ref=True | just_header=False
[...]

PR #1995, opened 2015-10-09
Update OneFichierCom.py

GammaC0de:
id is not needded anymore

### DIFF
--- a/module/plugins/hoster/OneFichierCom.py
+++ b/module/plugins/hoster/OneFichierCom.py
@@ -9,10 +9,10 @@ from module.plugins.internal.SimpleHoster import SimpleHoster, create_getInfo
 class OneFichierCom(SimpleHoster):
     __name__    = "OneFichierCom"
     __type__    = "hoster"
-    __version__ = "0.91"
+    __version__ = "0.92"
     __status__  = "testing"
 
-    __pattern__ = r'https?://(?:www\.)?(?:(?P<ID1>\w+)\.)?(?P<HOST>1fichier\.com|alterupload\.com|cjoint\.net|d(es)?fichiers\.com|dl4free\.com|megadl\.fr|mesfichiers\.org|piecejointe\.net|pjointe\.com|tenvoi\.com)(?:/\?(?P<ID2>\w+))?'
+    __pattern__ = r'https?://(?:www\.)?(?:\w+\.)?(?P<HOST>1fichier\.com|alterupload\.com|cjoint\.net|d(es)?fichiers\.com|dl4free\.com|megadl\.fr|mesfichiers\.org|piecejointe\.net|pjointe\.com|tenvoi\.com)(?:/\?\w+)?'
     __config__  = [("use_premium", "bool", "Use premium account if available", True)]
 
     __description__ = """1fichier.com hoster plugin"""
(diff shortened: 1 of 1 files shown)


thadrien:
Thanks GammaC0de. This point was not clear for me.

thadrien:
Hi,

Could someone explain to me why ID1 and ID2 where not in self.info['pattern'] although they were still in self.__pattern__ regular expression ?

Thanks,

GammaC0de:
Ok,
Thanks to your question, I found the root cause of this bug and I think of #1979 too :)
see #1997 for fix

many thanks :smile:

thadrien:
Thanks for your answer.
But, still don't understand something : why self.info['pattern'] was not
empty, and why they weren't any message about re module missing ?

GammaC0de:
first of all, there was no self.info['pattern'] at all (at least when I checked).
as for message about re module missing, the exception was catched and ignored,
see [Base.py#L118](https://github.com/pyload/pyload/blob/stable/module/plugins/internal/Base.py#L118)
```python
        try:
            info['pattern'] = re.match(cls.__pattern__, url).groupdict()
        except Exception:
            pass
```

thadrien:
Thanks for your answer.
It was here when I checked this. I even insert a print instruction which
has shown its contents.
Catching exceptions without logging is not very safe. I will add something
to avoid this.

GammaC0de:
Yes, I also thought so but the problem is that this is `@classmethod` function so there is no `self` variable and you can't use `self.log_error()`.
you cant use `print()`, but that wont be logged to the log file.
@vuolter is there any other method for logging that fits  to this case?

thadrien:
But what about exception filtering ?

vuolter:
@GammaC0de Maybe:
```
import logging
logger = logging.getLogger("log")
level = "error"
log = getattr(logger , level)
log("What you want is logged")
```

thadrien:
@vuolter : thx for your answer. Are you already using python logging module
in the other parts of the code ?

GammaC0de:
@vuolter how about adding a `pylog` function to `__builtin__` in v0.4.10 for these cases (this had already been done for `pyreq`)?

vuolter:
In 0.4.10 should be pycore, so you should able to call pycore.log directly.

GammaC0de:
:+1:  thats wonderful news :smile:

thadrien:
Thx so much vuolter. Hope it will come soon, at least a preview.

thadrien:
@vuolter : I think we should use one and only one logging system, because using two systems will make some sync problems between the two systems. Adding pycore to builtins seems a better solution to me, although I don't like global variables.

To avoid this problem to occur again, maybe we should filter exception, like this :
```python
except NameError:
  raise
except Exception:
  pass
```

GammaC0de:
@thadrien the problematic part of you suggestion is:
```
except Exception:
  pass
```
so unhandled exception will still be ignored (that was the case in the first place) and you have to carefully think what type of exceptions can occur.

thadrien:
@GammaC0de I totally agree with you. But at least it will avoid forgetting `import re` in the future.

thadrien:
Or something else :
```
import re
try:
    info['pattern'] = re.match(cls.__pattern__, url).groupdict()
except Exception:
    pass
```

If re module is already imported, `import re` will do nothing. Otherwise, it will import re. Since this import is closer to the point where it will be used, there are less risks of forgetting it again.

GammaC0de:
better is:
```python
except IndexError:
  pass
except Exception:
  raise
```

thadrien:
:+1:

PR #2015, opened 2015-10-12
Pathprefix Support

rix1337:
This pull is based largely on https://github.com/pyload/pyload/pull/736
and the Next theme https://github.com/pyload/pyload/pull/1006

I have never used git before and am also new to python.
Therefore please look for coding errors especially in `pyload_app` and `utils`.

So far the interface & path prefix setting work fine with some exceptions:
- Hardcoded `/pyload` prefix in `settings_item.html` because I couldn't escape the brackets right
- Can't add Links
- Favicon doesn't show up behind my reverse proxy, on direct connection it does (maybe a caching bug on my browser/cloudflare)
- Other templates untouched


EdVanDance:
Maybe the url jinja filter should be used to handle the prefix? The filter is defined in module/web/webinterface.py.

GammaC0de:
https://github.com/pyload/pyload/pull/3077

PR #2019, opened 2015-10-13
[Plugin] improve error tracing (2)

GammaC0de:
Hi,
I was not happy with the current tracing implementation - it does not print exception information
so I improved tracing:
* Now tracing goes to the log (with `log_exc()` function)
* `log_exc()` will log a `traceback.print_exc()` message with FULL traceback if it detects an exception, otherwise it will log `traceback.print_stack()` message.
* Tracing is off by default (except critical)
* `log_xxx(tracing=True)` will print the traceback regardless of debug mode or not (except `log_debug()`)

Should we print trace only in debug mode? I don`t have a definitive answer for this

### DIFF
--- a/module/plugins/internal/Plugin.py
+++ b/module/plugins/internal/Plugin.py
@@ -284,7 +284,7 @@ def chunks(iterable, size):
 class Plugin(object):
     __name__    = "Plugin"
     __type__    = "plugin"
-    __version__ = "0.57"
+    __version__ = "0.58"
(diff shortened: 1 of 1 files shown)


vuolter:
Great work! Maybe `log_exc()` could be moved to `_log` to simplify the code.
Yep, tracing should be active in debug mode only.

GammaC0de:
:+1: feel free to improve it..

PR #2021, opened 2015-10-13
Fixed sid storage pyload#2007

semicuda:


vuolter:
I'll push a proper fix later today. ;)

semicuda:
It worked :P But thanks for having a look, I'm curious about how to fix it properly

PR #2041, opened 2015-10-16
[internal/Accounts.py] Fix unintended password change

joberreiter:
Changing `safe_info['login']['password']` to mask the output of the password in the debug log also changes the original password in `self.info`!
A solution is to deepcopy `self.info` (a shallow copy is not sufficient).


joberreiter:
Maybe fixes a couple of problems regarding account logins ...

vuolter:
I directly removed the safe_info dict.

PR #2044, opened 2015-10-17
[Hoster] fix #2036

### LINKED ISSUE #2036: BasePlugin doesn't allow multiple connections per download
Hi All, I don't know whether it is a pyload problem or the problem providing direct download links to pyload but it seems that all downloads only use a single connection even though I have set 3 chunks per download - has anybody else noticed this?

Best
chaos

GammaC0de:
fix #2036


vuolter:
Already fixed.

PR #2053, opened 2015-10-18
[BasePlugin] Http hoster is not an internal hoster

jellysheep:
This fixes an import of the [Http hoster](https://github.com/pyload/pyload/blob/aff9388d433847fa7adad7421c4d76bd5bdd4e5a/module/plugins/hoster/Http.py), which is not an internal hoster.
I don't know if it is intended to move `Http.py` to the `internal` directory?


vuolter:
No, it's a typo. Thanks!

PR #2066, opened 2015-10-19
pyLoadGui: fix bad import 

runningbits:
The local GUI probably won't work anyway but at least the import is fine now.


vuolter:
No more GUI in 0.5.x

PR #2095, opened 2015-10-21
[SimpleCrypter] update

GammaC0de:


vuolter:
I should be more careful, but I do not have time. :(
Anyway, thank you so much for your help! :)

GammaC0de:
:+1:

PR #2114, opened 2015-10-25
Bug correction in setup.py: new versions of PIL doesn't provide Image but PIL.Image

thadrien:


vuolter:
No more setup in 0.5.x

PR #2127, opened 2015-10-28
[SimpleHoster] [SimpleCrypter] update

GammaC0de:


vuolter:
`get_info` is a classmethod and doesn't have `self`

Issue #2061, opened 2015-10-19
Security issue when displaying debug logs


joberreiter:
With the latest commit b0ddd95c9697c2df7a9cf26398406fb57d2c5059, the user account passwords are now displayed unmasked in the logs.

As this is ok for local debugging purposes, it is worse in terms of security, as the log had to be edited to prevent revealing the account credentials to the world.

Is this an intended change?
Although not very beautiful at coding, there was a good reason for password masking in the past ...

GammaC0de:
@joberreiter Just a note: this masking was not very effective since every HTTP request, including the arguments, is logged in debug mode..
[Hoster.py#L176](https://github.com/pyload/pyload/blob/stable/module/plugins/internal/Hoster.py#L176), [Plugin.py#L217](https://github.com/pyload/pyload/blob/stable/module/plugins/internal/Plugin.py#L217)
```python
    def load(self, url, get={}, post={}, ref=True, cookies=True, just_header=False, decode=True,
             multipart=False, redirect=True, req=None):
        """
        Load content at url and returns it

        :param url:
        :param get:
        :param post:
        :param ref:
        :param cookies:
        :param just_header: If True only the header will be retrieved and returned as dict
        :param decode: Wether to decode the output according to http header, should be True in most cases
        :return: Loaded content
        """
        if self.pyload.debug:
            self.log_debug("LOAD URL " + url,
                           *["%s=%s" % (key, val) for key, val in locals().items() if key not in ("self", "url", "_[1]")])

```

vuolter:
@GammaC0de has already exposed the reason that led me to remove the password masking.

PR #2150, opened 2015-11-03
[Plugin] censor ACCOUNT request data

GammaC0de:
HI,
After the third time I see someone post a log that reveals his user and password, I think that @joberreiter was right in #2061
To overcome the request logging issue discussed in #2061, this PR will mask the requests of all account plugins.

Now you can reimplement `safe_info`.

### DIFF
--- a/module/plugins/internal/Account.py
+++ b/module/plugins/internal/Account.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 
+import copy
 import random
 import re
 import threading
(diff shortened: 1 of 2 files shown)


joberreiter:
+1
`safe_info` was only a temporary variable of `self_info` with the password anonymized, so there is no need to reimplement it, as its only purpose---the hiding of password/keys in the logs---is already fulfilled with your fix.

GammaC0de:
Ok, added `safe_info` to the PR.

Maybe on v0.4.10 you should add a config parameter `mask sensitive information in debug log` so one can disable this cencorship when needed

GammaC0de:
Sorry, this is a major issue (see #2194)

PR #2164, opened 2015-11-14
added two old crypters

igel-kun:
I've been using these two for a while.


vuolter:
Warez website or services are not allowed in the official repository.

PR #2176, opened 2015-12-03
fixed Keep2Share redirection mess

igel-kun:
Keep2Share has many names: keep2share.cc, k2s.cc, keep2s.cc and if a link is for the wrong version, they will HTTP-redict to the correct one.

So, If I paste a link for k2s.cc/bla into pyload but the file is at keep2s.cc/bla. Then, k2s.cc will respond with 302, location: keep2s.cc/bla. That's fine as long as we don't want to POST anything, since, by default, libcurl will follow the sequence POST->302 with GET, and what we receive for our captcha submission is the original page, and pyload gets confused because there is no answer to our captcha submission.

Fix 1: one could  add
```
self.c.setopt(pycurl.POSTREDIR, pycurl.REDIR_POST_ALL)
```
to initHandle() in HTTPRequest.py - note that this requires libcurl 7.17.1 (see http://curl.haxx.se/libcurl/c/CURLOPT_POSTREDIR.html)

Fix 2: one could add a method following all redirects at the beginning of a process() and replace pyfile.url with the result of the last redirect.

In fact, the second option is preferable since the different keep2share variants seem to handle their captchas independently meaning, if k2s.cc gives you a captcha & you submit the (correct) solution to keep2s.cc, it will result in a "captcha invalid" response because it's not submitted to the right server.

I figured, Base.py would be the right place to implement this, as following redirects is not hoster specific but might apply to crypers too...

### DIFF
--- a/module/plugins/hoster/Keep2ShareCc.py
+++ b/module/plugins/hoster/Keep2ShareCc.py
@@ -10,7 +10,7 @@ from module.plugins.internal.SimpleHoster import SimpleHoster, create_getInfo
 class Keep2ShareCc(SimpleHoster):
     __name__    = "Keep2ShareCc"
     __type__    = "hoster"
-    __version__ = "0.28"
+    __version__ = "0.29"
     __status__  = "testing"
 
     __pattern__ = r'https?://(?:www\.)?(keep2share|k2s|keep2s)\.cc/file/(?P<ID>\w+)'
@@ -25,8 +25,6 @@ class Keep2ShareCc(SimpleHoster):
     __authors__     = [("stickell", "l.stickell@yahoo.it"),
                        ("Walter Purcaro", "vuolter@gmail.com")]
 
-
-    URL_REPLACEMENTS = [(__pattern__ + ".*", "http://keep2s.cc/file/\g<ID>")]
 
     NAME_PATTERN = r'File: <span>(?P<N>.+?)</span>'
     SIZE_PATTERN = r'Size: (?P<S>[^<]+)</div>'
@@ -115,7 +113,7 @@ class Keep2ShareCc(SimpleHoster):
         m = re.search(self.CAPTCHA_PATTERN, self.data)
         self.log_debug("CAPTCHA_PATTERN found %s" % m)
         if m is not None:
-            captcha_url = urlparse.urljoin("http://keep2s.cc/", m.group(1))
+            captcha_url = urlparse.urljoin(self.pyfile.url, m.group(1))
             post_data['CaptchaForm[code]'] = self.captcha.decrypt(captcha_url)
         else:
             recaptcha = ReCaptcha(self)
--- a/module/plugins/internal/Base.py
+++ b/module/plugins/internal/Base.py
@@ -36,7 +36,7 @@ def create_getInfo(klass):
 class Base(Plugin):
     __name__    = "Base"
     __type__    = "base"
-    __version__ = "0.19"
+    __version__ = "0.20"
     __status__  = "stable"
 
     __pattern__ = r'^unmatchable$'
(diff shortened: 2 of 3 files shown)


GammaC0de:
@igel-kun do you have a sample url that has this issue to test?

igel-kun:
I guess you can just take any keep2share link and change the prefix. I randomly picked this up from the internet (no idea about the content): http://keep2share.cc/file/61887b44e8e92
which should reproduce the issue as it's returning (for me) 302, location: http://k2s.cc/...

GammaC0de:
@igel-kun I think modifying `self.pyfile.url` is not a good practice as it can break other plugins.
Also, can you post a debug log with the problem?

igel-kun:
this is with the current plugin situation:
```
07.12.2015 10:57:47 DEBUG     HOSTER Keep2ShareCc[17871]: Plugin version: 0.28
07.12.2015 10:57:47 DEBUG     HOSTER Keep2ShareCc[17871]: Plugin status: testing
07.12.2015 10:57:47 WARNING   HOSTER Keep2ShareCc[17871]: Plugin may be unstable
07.12.2015 10:57:47 INFO      HOSTER Keep2ShareCc[17871]: Grabbing link info...
07.12.2015 10:57:47 DEBUG     HOSTER Keep2ShareCc[17871]: Link info: {'status': 2, 'pattern': {'S': u'8.0 MB', 'ID': u'<some ID>', 'N': u'<some name>'}, 'size': 8388608, 'url': u'http://keep2s.cc/file/<some ID>', 'name': u'<some name>'}
07.12.2015 10:57:47 DEBUG     HOSTER Keep2ShareCc[17871]: Previous link info: {}
07.12.2015 10:57:47 INFO      HOSTER Keep2ShareCc[17871]: Link name: <some name>
07.12.2015 10:57:47 INFO      HOSTER Keep2ShareCc[17871]: Link size: 8.00 MiB (8388608 bytes)
07.12.2015 10:57:47 INFO      HOSTER Keep2ShareCc[17871]: Link status: online
07.12.2015 10:57:47 INFO      HOSTER Keep2ShareCc[17871]: Looking for direct download link...
07.12.2015 10:57:47 DEBUG     HOSTER Keep2ShareCc[17871]: Redirect #0 to: http://keep2s.cc/file/<some ID>
07.12.2015 10:57:47 DEBUG     HOSTER Keep2ShareCc[17871]: LOAD URL http://keep2s.cc/file/<some ID> | redirect=True | cookies=True | get={} | req=None | decode=True | multipart=False | post={} | ref=True | just_header=True
07.12.2015 10:57:48 INFO      HOSTER Keep2ShareCc[17871]: Direct download link detected
07.12.2015 10:57:48 INFO      HOSTER Keep2ShareCc[17871]: Downloading file...
07.12.2015 10:57:48 DEBUG     HOSTER Keep2ShareCc[17871]: DOWNLOAD URL http://keep2s.cc/file/<some ID> | cookies=True | resume=None | get={} | disposition=True | chunks=None | post={} | ref=True
07.12.2015 10:57:48 INFO      HOSTER Keep2ShareCc[17871]: Checking downloaded file...
07.12.2015 10:57:48 WARNING   Download failed: <some name> | File size mismatch | Expected file size: 8388608 | Downloaded file size: 15050
```

and the downloaded file is the welcome-page of keep2share since ```Hoster::isdownload()``` receives a 302 and thus returns the link in the ```location``` field of the header. Then, ```SimpleHoster::handle_direct()``` returns ```pyfile.url``` which still points to the welcome-page making ```SimpleHoster::process()``` think this is a direct download...

I would agree that modifying pyfile.url is bad if it was saved back into the database (if the redirect was temporary, the correct URL might be lost), but I don't think it is. Do you know a plugin that might break from having pyfile.url modified according to a redirect that I'm proposing? Or maybe a theoretical scenario in a future plugin that might break?

PR #2228, opened 2015-12-29
 decryption was failing, with error Decrypting failed:****.dlc | global name 'pyfile' is not defined

ckaeslin:
dlc decryption was failing, with error Decrypting failed:****.dlc | global name 'pyfile' is not defined
debug mode showed missing self


vuolter:
THX :)

PR #2233, opened 2015-12-29
'super' object has no attribute 'check_download'

ckaeslin:
In earlier versions (before Octobre) there was a check_download in the Hoster.py but now there is only a _check_download.
Not sure what the idea is here, but with this change everything works fine and the download progress bar in queue is updated correctly.


vuolter:
`check_download` is a public method, `_check_download` is private. ;)
Typo fixed here: https://github.com/pyload/pyload/commit/f7d58afbccabb735bc5865222991f5b204f53725
Thanks for reporting! :)

PR #2263, opened 2016-01-04
[ClickNLoad] Shutdown gracefully + code cleanup

GammaC0de:
As the title says:
`ClickNLoad` will gracefully close the server socket befor exiting - no more "address already in use" errors (hopefully)

One question:
What is the purpose of the startup delay?


vuolter:
Sometimes addons are activated before the webserver was ready-

GammaC0de:
Ok, the startup delay is still there, can we merge this?

vuolter:
Sorry for the late, I was a little busy these days. ;)

Issue #2289, opened 2016-01-13
UnRAR plugin issues "Extract failed"


Arno-Nymous:
The UnRAR extractor (plus, as it seems to me, the other extractor plugins, too) reports an "Extract failed" error message but gives no further information. This error message can pass unnoticed since the traceback is not printed in `--daemon` mode. One has to run pyLoad in the foreground in order to see where the traceback leads.

Further examination revealed to me that as soon as the UnRAR plugin hits [this line](https://github.com/pyload/pyload/blob/stable/module/plugins/internal/UnRar.py#L227) it chokes on the upfollowing log commands since the value of `pyfile.plugin` can be `None`.

Unfortunatelly I don't have a log ouput at this very moment which demonstrates this behavior, but I'll do a pull request in no time, which fixes this error plus some more which went hand in hand with this one.

GammaC0de:
see #2250

PR #2290, opened 2016-01-13
Fixes the "Extract failed" messages in UnRAR extractor plugin

Arno-Nymous:
See [Issue 2289](https://github.com/pyload/pyload/issues/2289).

### DIFF
--- a/module/database/FileDatabase.py
+++ b/module/database/FileDatabase.py
@@ -21,11 +21,12 @@
 from threading import RLock
 from time import time
 
-from module.utils import formatSize, lock
 from module.PullEvents import InsertEvent, ReloadAllEvent, RemoveEvent, UpdateEvent
-from module.PyPackage import PyPackage
 from module.PyFile import PyFile
+from module.PyPackage import PyPackage
 from module.database import style, DatabaseBackend
+from module.utils import formatSize, lock
+
 
 try:
     from pysqlite2 import dbapi2 as sqlite3
@@ -831,7 +832,9 @@ class FileMethods():
         self.c.execute("SELECT url, name, size, status, error, plugin, package, linkorder FROM links WHERE id=?", (str(id), ))
         r = self.c.fetchone()
         if not r: return None
-        return PyFile(self.manager, id, * r)
+        pyfile = PyFile(self.manager, id, * r)
+        pyfile.initPlugin()
+        return pyfile
 
 
     @style.queue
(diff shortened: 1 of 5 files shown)


module/database/FileDatabase.py
```diff
@@ -831,7 +832,9 @@ def getFile(self, id):
         self.c.execute("SELECT url, name, size, status, error, plugin, package, linkorder FROM links WHERE id=?", (str(id), ))
         r = self.c.fetchone()
         if not r: return None
-        return PyFile(self.manager, id, * r)
+        pyfile = PyFile(self.manager, id, * r)
```
Arno-Nymous: Admittedly have to mention that I don't know whether this is the right place to re-init the `pyfile.plugin`.
Maybe there is a better place to put in order to remain DRY. But it's a fact that in some place there has to be made sure that `pyfile.plugin` should never be `None` but set to the appropriate plugin.

vuolter:
I'll have a look at this soon I can.

module/plugins/hooks/ExtractArchive.py
```diff
@@ -275,17 +275,17 @@ def extract(self, ids, thread=None):  #@TODO: Use pypack, not pid to improve met
             if not exists(out):
                 os.makedirs(out)
 
-            matched   = False
-            success   = True
-            files_ids = dict((fdata['name'], ((fsjoin(dl_folder, pypack.folder, fdata['name'])), fid, out)) for fid, fdata \
-                        in sorted(pypack.getChildren().values(), key=lambda k: k['name'])).items()  #: Remove duplicates
+            matched = False
+            success = True
+            files_ids = dict((fdata['name'], (fdata['id'], fsjoin(dl_folder, pypack.folder, fdata['name']), out)) for fdata \
```
Arno-Nymous: Fixes [#2296](https://github.com/pyload/pyload/issues/2296)

GammaC0de:
@Arno-Nymous Maybe your PR does fix the above issue, but it also changes fundamental parts of `pyLoad`'s core so I'm pretty sure that it will be rejected by @vuolter, never the less that it **was** working before (some changes were made to it that broke it) without the changes in the core.. :wink:

nordic-style:
@GammaC0de i did not check the pr but i would say that there are some core changes that are required to "fix" pyload in a proper way?! ;)

manuelm:
The PR is incomplete:
* The order of the list elements in https://github.com/pyload/pyload/pull/2290/files#diff-5bc8c590ca33f5f573f2a9570649a67dR348 needs to be adapted as well.
* `pyfile.plugin` gets released in https://github.com/pyload/pyload/pull/2290/files#diff-5bc8c590ca33f5f573f2a9570649a67dR326 but needs to stay alive until https://github.com/pyload/pyload/pull/2290/files#diff-5bc8c590ca33f5f573f2a9570649a67dR350

PR #2343, opened 2016-02-17
Avoiding UnboundLocalError when using method 'type' on method get_info from SimpleHoster.

YiFanChen99:
If someone use "type" method to verify the type of an object on get_info when debugging. They will be a "UnboundLocalError: local variable 'type' referenced before assignment" Error. So I recommend avoiding this name for a temporary variable.


vuolter:
Thanks! I really hate this kind of careless mistakes. :D

PR #2350, opened 2016-02-20
Fix configdir

### LINKED ISSUE #1557: Commandline Attributes not working
--configdir and --pidfile are not working properly at least not for me ... even in the 0.4.9 build

sodd:
this PR should fix #1557 and as a bonus i added argparser.


Velociraptor85:
@GammaC0de lets take a look at this ... seems good code I will try to verify / test this weekend ... also seems to work with 3.x and 2.7 at least what I understand from https://docs.python.org/2/library/argparse.html

PR #2358, opened 2016-02-23
[ReCaptcha] Overhauled handling of NoCAPTCHA ReCAPTCHA (ReCAPTCHA v2)

Arno-Nymous:
Can now send those clickable 3x3 images to user via webinterface or captcha services like 9kw.
Can be solved by typing the numbers of the accoring picture tiles.


Arno-Nymous:
Don't pull this change for now. I noticed two things a want to fix beforehand. I'll open a new pull request when finished.

PR #2374, opened 2016-03-06
V8 is not thread-safe. Make sure we lock it

manuelm:
This fixes an annoying crash when using the threaded CherryPyWSGIServer.


vuolter:
So, it was merged in https://github.com/pyload/pyload/pull/2450. Continue there, but thank you for that PR! :D

PR #2375, opened 2016-03-06
[UlozTo] Added support for audio Captcha recognition

sodd:
I would like to add this feature to the base ulozto source code. More info could be found at https://github.com/izderadicka/adecaptcha and I hope @izderadicka would not mind this PR.


vuolter:
Interesting.

PR #2439, opened 2016-04-27
Few plugin changes

Arno-Nymous:
[ShareLinksBiz] Speed up link parsing by cascading through all available container types.
~~[Captcha.py] Change string formatting to `.format()` here as well. Also fixes an `TypeError`
occuring on line 106.~~


GammaC0de:
:+1:

PR #2450, opened 2016-05-03
[JsEngine] Support large scripts

GammaC0de:
Currently every JS script is sent as a command line parameter.
The length of the command line is limited of-course (especially on windows)

This PR will make `pyLoad` use a temporary file as a script if the script size exceeds some specific length.
I've included #2374 in it also.

### DIFF
--- a/module/common/JsEngine.py
+++ b/module/common/JsEngine.py
@@ -17,9 +17,11 @@
     @author: RaNaN
 """
 
+import os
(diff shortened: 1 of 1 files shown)


vuolter:
Cherry-picked.

vuolter:
Closed because only Js2Py will used in 0.5 .

PR #2457, opened 2016-05-07
20160507 stable safelinking

tbsnpy:
- Adapted SafelinkingNet plugin to support new site logic
- Added SolveMediaJson plugin, which supports JSON interface of SolveMedia
- No selenium/phantomjs needed.


vuolter:
@tbsnpy Conflicts with https://github.com/pyload/pyload/pull/2455?

tbsnpy:
Yes, this plug-in is preferable to the SafelinkingNet plug-in from #2455, since it does not need selenium and phantomjs. 

#2455 additionally includes a ClicknUpload plug-in, which is why I have not closed that pull request, yet. I'm currently rewriting that plug-in without the selenium and phantomjs dependencies and will close #2455, when I'm finished.

GammaC0de:
Ok, so why post if its not finished yet?

tbsnpy:
You are referring to #2455, I guess? At the point of the pull request I was not sure, if I'd be able to come up with a version without the selenium/phantomjs dependencies.

GammaC0de:
No I was referring to this PR (#2457).

Another question, what is SolveMediaJson.py? why to use it and not just use SolveMedia.py?

tbsnpy:
The SafelinkingNet plug-in from this pull request is finished. The unfinished plug-in I was talking about is the ClicknUpload plug-in without the selenium/phantomjs dependencies.

I added the SolveMediaJson.py captcha, because I did not want to change the existing SolveMedia.py and possibly break existing plug-ins that use the original version. Safelinking.net uses JSON to determine the SolveMedia captcha. 

The current SolveMedia.py plug-in is not using JSON but regular GET/POST requests, as far as I can tell. An example: The SolveMedia challenge is parsed from the input field 'adcopy_challenge'. This input field is not available, when loading Safelinking.net via pycurl.

I could extend the existing SolveMedia interface to alternatively use JSON instead, I you'd prefer it that way.

GammaC0de:
Extending existing SolveMedia.py to handle both cases would be preferable, no need for duplicates if you can avoid them...

tbsnpy:
Agreed. I'll implement the necessary changes.

tbsnpy:
The captcha retrieval works with the existing plug-in SolveMedia.py as well. :+1: 

* Adapted SafelinkingNet plug-in
* Removed custom plug-in SolveMediaJson.py

GammaC0de:
@tbsnpy Please post sample URL with captcha, tnx

tbsnpy:
@GammaC0de Please see the links below:

http://safelinking.net/L9rPths (no password)
http://safelinking.net/12ketqB (password: test)

The plug-in retrieves the package password with 'self.get_password()'. I've seen that other plug-ins use that function as well.

The returned password is the one the user has entered in the 'Password - Password for RAR-Archive' field of the "Add package" dialogue. The field name is pretty misleading and I assume that I'm using the wrong interface.
Is there another way to enter passwords for password protected links?

GammaC0de:
:+1: 
No, currently this is the proper way.

PR #2508, opened 2016-06-14
Oboom - constantly asking for solving captcha

OzzieIsaacs:
As free user Oboom is constantly asking for solving captchas, even if captcha was solved correctly


GammaC0de:
:+1:

PR #2512, opened 2016-06-19
[NitroFlareCom] Use NitroFlare API for Premium to reduce CAPTCHA issues.

anonymous-aardvark:
Switch to the NitroFlare API to reduce CAPTCHA issues.  Not sure if the API ever uses CAPTCHAs of any sort so may need adjusting if people to encounter CAPTCHA issues.

Code was bashed out at 4am so probably not perfect.  First time working with pyload plugins - sorry if I missed anything, but it works for me so far.


anonymous-aardvark:
Version bump not included - both account and hoster plugins should probably be bumped if this is merged.

GammaC0de:
:+1: https://github.com/pyload/pyload/commit/79dad368430c24a89fd2502dddb7f1b2bb012a71

PR #2539, opened 2016-07-11
Fixed hoerbuchIn plugin

EikeKre:
Hoerbuch.in moved to hoerbuch.us.

### DIFF
--- a/module/plugins/crypter/HoerbuchIn.py
+++ b/module/plugins/crypter/HoerbuchIn.py
@@ -1,19 +1,16 @@
 # -*- coding: utf-8 -*-
 
 import re
-
 import BeautifulSoup
-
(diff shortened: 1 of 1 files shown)


module/plugins/crypter/HoerbuchIn.py
```diff
@@ -35,30 +33,42 @@ def decrypt(self, pyfile):
...
-        m = self.protection.search(url)
+
+        m = self.hoster_links.search(url) or self.protection.search(url)
+
         if m is None:
             self.fail(_("Bad URL"))
         url = m.group(0)
 
-        self.pyfile.url = url
+        if self.hoster_links.match(url):
```
EikeKre: This may be done better as I only want the final url from the referral url http://www.hoerbuch.us/wp/goto/Download/34534564

PR #2544, opened 2016-07-14
feature: delete multiple links by hoster

EikeKre:
When a package opens a select box is created with the options to select all, none or by hoster. If any link is checked a buttons appears to delete all selected links.


vuolter:
I'll take a look next week!
...but what is the reason why it's closed?

EikeKre:
@vuolter thanks for your reply. I accidentally deleted the branch. Should I restore it?

PR #2545, opened 2016-07-14
Fix unpack zipfiles

EikeKre:


vuolter:
Merged but better implementation wanted. Anyway thank you always! ;)

PR #2553, opened 2016-07-22
keep2share redirect fix

igel-kun:
there is a problem with keep2share forwarding to its other domains in combination with captchas. If a file is asked from keep2s.cc and we're being redicted to k2s.cc, then the captcha request still goes to keep2s.cc instead of k2s.cc so even if we solve it correctly, the server will say its invalid, because we're sending it to the wrong server.
In this patch, I'm following all redirects before requesting the file.


igel-kun:
Oh, I also moved the parsing of the time in format "HH:MM:SS" to parse_time in internal/misc.py

GammaC0de:
same reasone as https://github.com/pyload/pyload/pull/2562#issuecomment-236577317

Issue #2402, opened 2016-03-30
Request for smule.com


psycholyzern:
I want to download the video on smule.com
Seems that they use javascript to decode the encoded url.

I see this guys, managed to use the same decoding method using javascript too, http://thebluebyte.com/getsmule/home/template/javascript/sutils.js

I can't understand what that script do in order to decode the link, so I'm asking if you can make this as plugin for pyload?

This is the url sample of a video https://www.smule.com/recording/eyshila-nada-pode-calar-um-adorador-eyshila/443694955_179520763

igel-kun:
The code I committed in pull request #2562 downloads the sample file for me.

GammaC0de:
:+1:

psycholyzern:
Thank you so much :)

PR #2562, opened 2016-07-31
smule hoster plugin, using updated JsEngine code (see Issue #2402)

igel-kun:
This "hoster" needs some javascript evaluated, so I took the liberty of updating ```/modules/common/JsEngine.py``` as it wasn't supporting the js engine I run (node_js)

### DIFF
--- a/module/common/JsEngine.py
+++ b/module/common/JsEngine.py
@@ -20,7 +20,7 @@
 from imp import find_module
 from os.path import join, exists
 from urllib import quote
-
+import subprocess
 
 ENGINE = ""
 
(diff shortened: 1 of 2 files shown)


GammaC0de:
Hi, In 'stable' branch, you cannot change anything besides "plugins" directory.

Also for changes in 'internal' directory you should create a separate PR because it effects *all* the plugins so @vuolter has to approve it.

igel-kun:
So, what do I do with the change to ```JsEngine.py``` which the hoster implementation depends on?

GammaC0de:
@igel-kun Isn't the current implementation of `JsEngine.py` enough?

igel-kun:
the current implementation forces ```eval(unescape('<script>'))``` on the script, which is OK for simple scripts but not OK for more complicated stuff containing variable and function declarations (so I added ```eval_raw()```). Secondly, it doesn't support ```Node JS``` which I happen to run, as it does not have a ```print``` command, but it uses ```console.log``` (so I added detection for ```Node JS```).
If I had left the current infrastructure unchanged, these two changes would have incurred silly code, so I had to change some supporting code.

Here's the problem with the old engine:
```
$ python2 -m module.common.JsEngine
  File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "module/common/JsEngine.py", line 162, in <module>
    js.eval(test)
  File "module/common/JsEngine.py", line 101, in eval
    raise Exception("No JS Engine")
Exception: No JS Engine
$ js -e 'console.log(23+19)'
42
$ file /usr/bin/js /etc/alternatives/js
/usr/bin/js:          symbolic link to /etc/alternatives/js
/etc/alternatives/js: symbolic link to /usr/bin/nodejs
```

Okay, now that I think of it, maybe the ```eval_raw()``` can be emulated with a bit more javascript-fu, which I don't have...

GammaC0de:
- What platform are you running?
- Do you have sample link to test this?

degs01:
rpi2 rasbian jessy 
no link

igel-kun:
same configuration here, a sample link can be found in the original issue #2402 I think

igel-kun:
ah nice, thanks for the adaptation! This additional ```get_info``` to add ".mp4" will also be necessary for #2561 I'm afrait... I'll update that PR later on...

GammaC0de:
:+1: 

Sorry for the dumb question, but what exactly is ARD?

igel-kun:
It's not at all a dumb question! ARD is one of Germany's public broadcasters. As their broadcasting is public, they keep the most recent shows around in what they call "medithek" for anyone (an IP registered to a German ISP might be required tho, not sure) to watch in case they missed it. See https://en.wikipedia.org/wiki/ARD_(broadcaster)

PR #2563, opened 2016-08-01
update parse_time in misc.py

igel-kun:
added functionality to parse times in format ```HH:MM:SS``` and ```MM:SS```

### DIFF
--- a/module/plugins/internal/misc.py
+++ b/module/plugins/internal/misc.py
@@ -592,6 +592,12 @@ def parse_time(value):
     if re.search("da(il)?y|today", value):
         seconds = seconds_to_midnight()
 
+    elif re.search("\d:\d\d", value):
+        # use the HH:MM:SS format, NOTE: when only one ':' is found, it assumes MM:SS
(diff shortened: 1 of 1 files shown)


igel-kun:
oh I should mention that keep2share uses this time format to diplay wait times.

module/plugins/internal/misc.py
```diff
@@ -592,6 +592,12 @@ def parse_time(value):
     if re.search("da(il)?y|today", value):
         seconds = seconds_to_midnight()
 
+    elif re.search("\d:\d\d", value):
+        # use the HH:MM:SS format, NOTE: when only one ':' is found, it assumes MM:SS
```
GammaC0de: Why assume MM:SS and not HH:MM?

igel-kun: Well, there is no way around assuming one of the two without context, and seeing that I stumbled upon that while handling wait times (they were always below 1h), this seemed the obvious choice. However, to be more general, maybe an argument should be added to the function telling it how to interprete a single colon.

PR #2578, opened 2016-08-27
Pyload Chef cookbook reference to README

### LINKED ISSUE #2574: Update domain in Readme
The ReadMe still has the old pyload.org domain. I planned to just do a PR with the domain names replaced. However the pages docs.pyload.net and pyload.net/configuration don't exist on the new domain, so I wasn't sure what to update them with.

djessich:
In past three weeks I have created a Pyload Chef Cookbook available here https://github.com/gridtec/cookbook-pyload. It is tested on different platforms, such as Debian, Ubuntu, Redhat, CentOS, Fedora and OpenSUSE and it supports ArchLinux. It is testd on [TravisCI](https://travis-ci.org/gridtec/cookbook-pyload). It is already released in version 1.0.0 to the [Chef Supermarket](https://supermarket.chef.io/cookbooks/pyload).

As I think that the cookbook should be referenced in the Pyload project, I added a reference to README. While I was working on that, I've also rewritten the README to markdown style and updated the domain in it (would also fix #2574).


Burnout5151:
This cookbook-approach is new to me. Sounds interesting.
Is there an advantage over DEB/RPM-Packages? Does it work for Windows/MacOS?

djessich:
Well, Chef its another way of doing things. A DEB or RPM package installs a package only for one platform (Debian or RedHat in that case) in one specific version and that's it. Thee packages inner layout is not equal, its different for different OS, so you need to create packages for each platform.

Whereas Chef focuses on whole infrastructure automation, a Chef cookbook offers the possibility to automatically install and maintain whatever the cookbook does (in this case install and maintain Pyload) for multiple OS (if the cookbook supports multiple OS) using a single source. You can decide if you like to just execute the cookbook once to install Pyload, or to execute it occasionally to install Pyload and maintain the installation, which means "hold it in that state, it has been described in the Chef cookbook". The latter would mean, if something is changed (f.e. the sys-v-init script), it will be reset by the Chef cookbook into the original state.

Moreover Chef provides you the ability to do standard system operations together with usual development standards (so called DevOps approach), such as testing (f.e. does a configfile has the content it should have?), version management, etc. With a single (or multiple) bash scripts this _may_ also work, but f.e. testing in bash is rather limited and also it is not OS transparent (see above), as you can see with the provided sys-v-init scripts in the Pyload cookbook.

More info on Chef can be found in the official documentation at https://www.chef.io/.

To answer your second question: Chef ships with Windows and MacOS support, but the cookbook I've written does not support them. Well, thats not the major problem, as we could implement the cookbook also for these platforms to install Pyload.

djessich:
I've worked on some improvements of the cookbook and I've added support for FreeBSD 10+. So the cookbook will work now on the following platforms: ArchLinux, Debian, Ubuntu, CentOS, RedHat, Fedora, OpenSUSE and FreeBSD.

djessich:
If @vuolter declares version 1.0.0 useable, I will make the cookbook compatible to it as soon as possible. Could we also reference the Pyload Chef cookbook in version 1.0.0 after its compatible?

PR #2600, opened 2016-09-18
[HTTPRequest] fix #2599

### LINKED ISSUE #2599: support for unofficial HTTP codes
see https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#Unofficial_codes
Currently, `__init__` of `BadHeader` crashes when receiving one of the unofficial codes, for example from Cloudflare.

GammaC0de:
fix #2599

### DIFF
--- a/module/network/HTTPRequest.py
+++ b/module/network/HTTPRequest.py
@@ -35,11 +35,26 @@ def myurlencode(data):
     return urlencode(dict((x.encode('utf_8') if isinstance(x, unicode) else x, \
         y.encode('utf_8') if isinstance(y, unicode) else y ) for x, y in data.iteritems()))
 
-bad_headers = range(400, 404) + range(405, 418) + range(500, 506)
+bad_headers = range(400, 404) + range(405, 600)
+
+unofficial_errors = {440: "Login Timeout - The client's session has expired and must log in again.",
+                     449: "Retry With - The server cannot honour the request because the user has not provided the required information",
+                     451: "Redirect - Unsupported Redirect Header",
+                     509: "Bandwidth Limit Exceeded",
+                     520: "Unknown Error",
+                     521: "Web Server Is Down - The origin server has refused the connection from CloudFlare",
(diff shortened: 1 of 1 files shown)


igel-kun:
may I humbly suggest the following:
```python
bad_headers = range(400, 404) + range(405, 600)

unofficial_errors = {440: "Login Timeout - The client's session has expired and must log in again.",
                     449: "Retry With - The server cannot honour the request because the user has not provided the required information",
                     451: "Redirect - Unsupported Redirect Header",
                     520: "Unknown Error",
                     521: "Web Server Is Down - The origin server has refused the connection from CloudFlare",
                     522: "Connection Timed Out - CloudFlare could not negotiate a TCP handshake with the origin server",
                     523: "Origin Is Unreachable - CloudFlare could not reach the origin server",
                     524: "A Timeout Occurred - CloudFlare did not receive a timely HTTP response",
                     525: "SSL Handshake Failed - CloudFlare could not negotiate a SSL/TLS handshake with the origin server",
                     526: "Invalid SSL Certificate - CloudFlare could not validate the SSL/TLS certificate that the origin server presented"}

class BadHeader(Exception):

    def __init__(self, code, content=""):
        int_code = int(code)
        Exception.__init__(self, "Bad server response: %s %s" % (code, httplib.responses[int_code] if int_code in httplib.responses else unofficial_errors[int_code]))
        self.code = code
        self.content = content

```

GammaC0de:
your code has bug.

it should be:
`Exception.__init__(self, "Bad server response: %s %s" % (code, httplib.responses[int_code] if int_code in httplib.responses else unofficial_errors.get(int_code, "unknown error code")))`

or better:
`Exception.__init__(self, "Bad server response: %s %s" % (code, httplib.responses.get(int_code, unofficial_errors.get(int_code, "unknown error code"))))`

Anyway,
I think @vuolter should decide which version to include.

igel-kun:
yes, good improvement :)

GammaC0de:
Merged 👍

PR #2621, opened 2016-10-16
more robust Fix for #2608

sebdelsol:
Sometimes Zippyshare uses other inline JS scripts to init document's attributes or JS vars that are used in the creation of the file url.

If this happens again as it has before, the plugin will stop to work.

So we need a robust way to catch the bad JS scripts, yet have all the valid ones still evaluated.


sebdelsol:
What's the issue?

PR #2622, opened 2016-10-16
Even more robust Fix for #2608

sebdelsol:
We need to evaluate _all_ JS scripts _in the same context_, except those that have _syntax errors_.

Can't test it at the moment, but it should work with the default js engine and PYV8.

### DIFF
--- a/module/plugins/hoster/ZippyshareCom.py
+++ b/module/plugins/hoster/ZippyshareCom.py
@@ -62,11 +62,19 @@ class ZippyshareCom(SimpleHoster):
         if self.link and pyfile.name == "file.html":
             pyfile.name = urllib.unquote(self.link.split('/')[-1])
 
-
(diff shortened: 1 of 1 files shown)


module/plugins/hoster/ZippyshareCom.py
```diff
@@ -62,11 +62,19 @@ def handle_free(self, pyfile):
         if self.link and pyfile.name == "file.html":
             pyfile.name = urllib.unquote(self.link.split('/')[-1])
 
-
+    def noSyntaxError(script) :
```
GammaC0de: missing 'self' parameter

GammaC0de:
After applying this PR, I had to apply #2450 also because of the **very** large JS script.
Even though, it still failing with `pycurl error 6`.

Issue #1616, opened 2015-07-30
UpdateManager doesn't keep settings


Stufo76:
UpdateManager plugin doesn't keep settings after pyLoad restart

Settings changed:

```
UpdateManager - "UpdateManager":
        bool autorestart : "Auto-restart pyLoad when required" = True
        int checkinterval : "Check interval in hours" = 1
        bool checkperiod : "Check for updates periodically" = True
        bool activated : "Activated" = True
        bool nodebugupdate : "Don't update plugins in debug mode" = False
        bool reloadplugins : "Monitor plugin code changes in debug mode" = True
        bool checkonstart : "Check for updates on startup" = True
```

Settings reverted back on pyLoad restart:

```
UpdateManager - "UpdateManager":
        bool autorestart : "Auto-restart pyLoad when required" = True
        int checkinterval : "Check interval in hours" = 8
        bool checkperiod : "Check for updates periodically" = True
        bool activated : "Activated" = True
        bool nodebugupdate : "Don't update plugins in debug mode" = False
        bool reloadplugins : "Monitor plugin code changes in debug mode" = True
        bool checkonstart : "Check for updates on startup" = True
```

Debian GNU/Linux 7.8 (wheezy)
3.2.0-4-amd64
Python 2.7.3

Velociraptor85:
@Stufo76 is this still the case? could this be a problem with file/folder permissions?

GammaC0de:
Fixed with https://github.com/pyload/pyload/commit/6c5c72b1a657e91f4ffee88629a3523cda8cacd1

PR #2625, opened 2016-10-18
Fix ConfigParser file permission issue

2chilled:
Fixes an issue which occurs when process owner has no write permission for `join(pypath, "module", "config", "default.conf")`. The later call to `open("pyload.conf", "wb")` then fails. Catched when executing pyLoadCore.py without an existing pyload.conf on a nix derivation (see http://nixos.org/) of pyload.

### DIFF
--- a/module/ConfigParser.py
+++ b/module/ConfigParser.py
@@ -58,6 +58,7 @@ class ConfigParser:
         try:
             if not exists("pyload.conf"):
                 copy(join(pypath, "module", "config", "default.conf"), "pyload.conf")
+                chmod("pyload.conf", 0600)
 
             if not exists("plugin.conf"):
                 f = open("plugin.conf", "wb")


2chilled:
Any chance to get this merged soon? This blocks pyload from being available on NixOS

GammaC0de:
@2chilled I'm afraid this can't be merged to the current release as no changes to the core are allowed. 
The next version of pyLoad (currently in development) should include this.

Velociraptor85:
is this the same like https://github.com/pyload/pyload/issues/1616 an thus also closed via the same commit?

GammaC0de:
👍 +1

FlorianFranzen:
@GammaC0de Why do I have to follow #3283 then if there are no changes to core. This whole project has become such a mess.

GammaC0de:
@FlorianFranzen The policy of "no changes to the core" is no more long time ago.
The above issue has already been fixed and also many more changes and bugfixes to the core, have been merged.
That is why you should follow #3283

PR #2677, opened 2017-01-06
Fix unrar

Georgeto:
Fixed a few issues with the ExtractArchive/UnRar plugin. Mainly interesting for the "set permissions" and "extract archive in archive" features, because they suffer from the incomplete unrar file listings.


vuolter:
Thank you! <3

PR #2683, opened 2017-01-11
[UploadedTo] Display correct valid until date even if expiration is near

Georgeto:
Display correct valid until date even if expiration is less than an hour away.


vuolter:
👍

Issue #2584, opened 2016-09-03
[v0.5] Questions


GammaC0de:
@vuolter Two questions:
- When we can see a beta version of v0.5?
- Where we can issue pull requests for v0.5?

kvoit:
Well the Twitter account is as silent as this bug tracker:
https://twitter.com/pyload
Two month ago, code was announced for "in a few weeks", news were "incoming" ... and then, questions have not been answered for the last month.

Roburetto:
Although a couple of strong competitors do apparently exist, the market shift and increased adoption of NAS and miniaturized Arm based devices such as Raspberry, OrangePi etc... gives PyLoad's Python architecture a decisive advantage.

wastez:
I hope it will not end like the first 0.5 Version a few years ago, which had a beta and then .... nothing.
Would be really great if pyload would get back to his old strength.
Hope the new version will come soon.
Thanks for your hard work.

Velociraptor85:
@vuolter 
+1 ^^

GammaC0de:
@vuolter is offline since July :cry:

wastez:
So pyload is dead again??

GammaC0de:
@wastez I cannot tell, hopefully @vuolter will return.
Meanwhile I am taking care of plugin issues.

Jonn3y:
Pyload has to be alive again! :(

GammaC0de:
@Jonn3y I've got a message  from @vuolter few days ago.
He is now back and he should resume his work soon (that's what he promised)

wastez:
Are there any news?
Is voulter again working on it?

vuolter:
@wastez I just have a few posts to read before do it...
![notifications](https://cloud.githubusercontent.com/assets/1221092/19415697/5bf4c41a-9378-11e6-942e-0930e2c52275.png)
:weary:

BKPepe:
Glad to see that @vuolter  is back and also alive!

Velociraptor85:
Happy to see @vuolter back here :+1:

godfuture:
Welcome back @vuolter

vuolter:
Thank you all.

GammaC0de:
@vuolter You should post something on Twitter..

Ryan747800B:
Welcome back @vuolter !

m33cr0b:
I cant wait to see this beta, since multi hoster support is broken, pyload has just become worthless to me!

[...]

PR #2717, opened 2017-02-04
[Click N' Load] Fix #2451

### LINKED ISSUE #2451: Names from site like filecrypt
Hello, since today or yesterday i see a small bug.
If i have load some files from filecrypt to pyload via click n load i get the folder name from filecrypt in pyload. looks like "xxx.1080p.x264-hermes -filecrpyt.cc". Now, i get only "filecrypt.cc". Where is the rest of the folder name?
I have also tested old files and have the same error.

GammaC0de:
Fix #2451

### DIFF
--- a/module/web/cnl_app.py
+++ b/module/web/cnl_app.py
@@ -69,7 +69,7 @@ def addcrypted():
 @local_check
 def addcrypted2():
 
-    package = request.forms.get("source", None)
+    package = request.forms.get("package", request.forms.get("source", None))
     crypted = request.forms["crypted"]
     jk = request.forms["jk"]
 


Velociraptor85:
will this be auto updated? since there is no version tag in this ? @GammaC0de

GammaC0de:
@Velociraptor85 It's not a plugin so UpdateManager cannot update it
It's a part of the core so it cannot get to the 'stable' branch (no changes to the core in 'stable' rule)
it has to be part of the next release.

Mister-XY:
Ok, when come the next release?

Velociraptor85:
@Mister-XY please check https://github.com/pyload/pyload/issues/2584

PR #2732, opened 2017-02-20
Download file size fix

dischinator:
I had a download report 1,000.00 as file size (I think it was filefactory) which was changed to 1.000.00 in misc.py causing an error when casted to float.
This patch checks if the size string contains a comma and a decimal point in that order and if so, the comma will be removed allowing the float cast to succeed.
The patched code will work as before if the string does not contain comma and decimal point.


dischinator:
Ah this looks even better thanks @GammaC0de

PR #2734, opened 2017-02-21
[MegaDebridEu] Remove plugin in useragent for the 404 error

kfranquenouille:
As I can see with the administrator of MegaDebridEu, the parameter `plugin="UserAgentSwitcher"` makes some problems. Natively, we can remove this param and use directly this class without using the plugin "UserAgentSwitcher"


GammaC0de:
User agent is configured via `UserAgentSwitcher` plugin configuration.

PR #2742, opened 2017-02-24
Fix several plugins

vuolter:
Debugged through `pylint -E`.


vuolter:
With the MultishareCz commit I completed this plugins revision! 😀

PR #2820, opened 2017-04-13
Update RealdebridCom.py

salimimani:
It seems that realdebrid links generated with the torrent service are of type : "https://real-debrid.com/d/{id}"
So the "**l**" in https://real-debrid.com/dl/" is not mandatory.
I rendered the **l** optional by replacing /**dl**/ with /**dl?**/ in order to match torrent service links.


GammaC0de:
Increase version please

PR #2831, opened 2017-04-23
Crypter Relink.us Update

OzzieIsaacs:
relink.us no longer worked. 
The password script moved to password2.php and to get to the real site 2 redirects have to be followed

### DIFF
--- a/module/plugins/crypter/RelinkUs.py
+++ b/module/plugins/crypter/RelinkUs.py
@@ -106,7 +106,7 @@ class RelinkUs(Crypter):
         self.package = pyfile.package()
 
     def request_package(self):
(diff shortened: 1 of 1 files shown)


module/plugins/crypter/RelinkUs.py
```diff
@@ -106,7 +106,7 @@ def init_package(self, pyfile):
         self.package = pyfile.package()
 
     def request_package(self):
-        self.data = self.load(self.pyfile.url)
+        self.data = self.load(self.pyfile.url, redirect=2)
```
GammaC0de: Why `redirect=2`?

OzzieIsaacs:
When I tried yesterday, it seemed to be necessary. I removed it.

PR #2836, opened 2017-04-29
Fix encoding issues when reading README

sebix:
on systems without UTF-8 as default (e.g. in build environments)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ from setuptools.command.sdist import sdist
 def _extract_text(path, fromline=None, toline=None):
     text = None
     path = os.path.join(os.path.dirname(__file__), path)
-    with io.open(path) as fp:
+    with io.open(path, encoding='UTF-8') as fp:
         if fromline or toline:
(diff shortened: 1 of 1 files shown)


vuolter:
@sebix Hi, `io.open` calls `locale.getpreferredencoding` with flag `do_setlocale=False` to determine the correct text encoding. If the result is undetermined we may forced the encoding to utf8, but we are really sure is always the right choice?! Could post more info aboit the issue you encountered. Thanks.

sebix:
The files are UTF-8 because of some chars (for the readme it's the the
copyright-char at the end). If the current locale is not utf-8, the
execution fails. Systems with desktop environments *may* always UTF-8
locales, but IMHO you shouldn't rely on this setting.
I encountered this issue when building packages using obs[1], which has
ANSI_X3.4-1968 (but it's actually irrelevant which actual encoding is used).

http://openbuildservice.org/

sebix:
Thanks!

vuolter:
@sebix Thanks to you for the PR! 😄

PR #2881, opened 2017-07-04
[DownloadScheduler] update timetable on change

adamantike:
### Type of request:

- [ ] Bugfix
- [ ] Code cosmetics
- [ ] Feature enhancement
- [ ] New feature
- [ ] New plugin

### Short description:

This triggers a schedule update every time the `timetable` config value is modified, instead of having to deactivate and activate the plugin again.

The only way I found to trigger a schedule update when I wanted to edit the `timetable` value, was by turning off and on the plugin, or waiting for it to trigger an update itself, based on the next time it determines to check the timetable again.

### DIFF
--- a/module/plugins/hooks/DownloadScheduler.py
+++ b/module/plugins/hooks/DownloadScheduler.py
@@ -9,7 +9,7 @@ from ..internal.Addon import Addon
 class DownloadScheduler(Addon):
     __name__ = "DownloadScheduler"
     __type__ = "hook"
-    __version__ = "0.27"
+    __version__ = "0.28"
     __status__ = "testing"
 
     __config__ = [("activated", "bool", "Activated", False),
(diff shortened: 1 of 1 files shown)


GammaC0de:
👍 Your PR + cosmetic changes.

adamantike:
@GammaC0de why did you need to create a different commit, add irrelevant changes to it, create a bug and merge it to a stable branch?

I'm willing to help and improve these requests accordingly, based on comments and a conversation with interested people. Please, consider following that practice, so contributors feel appreciated and contribute even more.

Also:
* If `activate` is executed before `config_changed`, `update_schedule` will be executed twice because we are now cleaning the last timetable here: https://github.com/pyload/pyload/blob/stable/module/plugins/hooks/DownloadScheduler.py#L25
* Handling translations this way is completely wrong: https://github.com/pyload/pyload/blob/stable/module/plugins/hooks/DownloadScheduler.py#L72. Not every language can negate a sentence by placing a negation word in it

GammaC0de:
@adamantike First of all, please don't be offended by my actions, my intentions were sincere:
I looked athe the code of DownloadScheduler.py and then I spotted some code cosmetics I liked to apply:
Instead of merging your PR and then adding my changes  (and so increasing the plugin version twice), I decided to do it in one commit.

> If activate is executed before config_changed, update_schedule will be executed twice because we are now cleaning the last timetable here: https://github.com/pyload/pyload/blob/stable/module/plugins/hooks/DownloadScheduler.py#L25

`activate()` function of a plugin acts like a constructor for the plugin, it will be run only once upon plugin initialization.
So it's a good place to initialize variables.
`config_changed()` is called only when you change the configuration via the WebUI, it will not be triggered on startup.

So, for my opinion, there is no problem here.
I also checked and `update_schedule()` indeed runs only once.



> Handling translations this way is completely wrong: https://github.com/pyload/pyload/blob/stable/module/plugins/hooks/DownloadScheduler.py#L72. Not every language can negate a sentence by placing a negation word in it.

You are right about this but:
- I did not write this plugin.
- I am sure that more stuff like this exists in (many?) other places.

You are very welcome to issue another PR to fix that if you like.

adamantike:
Thanks for the clarification, I appreciate it. I'll be creating a new PR and including the translation issue

PR #2914, opened 2017-07-27
[master] Avoid name mangling on pyload core object

adamantike:
### Type of request:

- [x] Bugfix
- [ ] Code cosmetics
- [ ] Feature enhancement
- [ ] New feature
- [ ] New plugin

### Short description:

Making the pyload core object a private (double-underscode) attribute `__pyload` is not a good idea because of name mangling when dealing with class inheritance.

### Additional info:

This means the private attribute is prepended with the base class' name, which means that, e.g. FileManager doesn't found its `__pyload` attribute because it's stored as `_BaseManager__pyload` instead.

For more information, see:
https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references

This could be fixed by using the `get_core` method instead, but I implemented a `pyload_core` property instead, to maintain the attribute's privacy and remove the getter.

Relevant log lines:
```
2017-07-27 08:22:54  CRITICAL  'FileManager' object has no attribute '_FileManager__pyload'
```

### DIFF
--- a/src/pyload/core/api/account.py
+++ b/src/pyload/core/api/account.py
@@ -22,7 +22,7 @@ class AccountApi(BaseApi):
 
         :return: string list
         """
-        return list(self.__pyload.pgm.get_plugins("account").keys())
+        return list(self.pyload_core.pgm.get_plugins("account").keys())
 
     @requireperm(Permission.Accounts)
     def get_accounts(self):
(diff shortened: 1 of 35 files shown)


adamantike:
@vuolter if you have time, this one is the top priority right now, as some new fixes collide with lines affected by this PR.

vuolter:
Merged but I'll rename `pyload_core` to `pyload` when I come back to the code.

adamantike:
I used that name, as something very simple would be a nightmare to grep actually.

Issue #2766, opened 2017-03-02
pycurl error 23: Failed writing body - UploadedTo


bbosiljcic:
I have Issues downloading from uploaded. 

I tried updating the hoster plugin, didn't help.

This is my log output. 

```02.03.2017 17:13:41	INFO	ACCOUNT UploadedTo: Grabbing account info for user 9*********...
02.03.2017 17:13:41	INFO	HOSTER UploadedTo[3]: Grabbing link info...
02.03.2017 17:13:41	INFO	HOSTER UploadedTo[3]: Link name: somefile.tar
02.03.2017 17:13:41	INFO	HOSTER UploadedTo[3]: Link size: 1.36 GiB (1464962860 bytes)
02.03.2017 17:13:41	INFO	HOSTER UploadedTo[3]: Link status: online
02.03.2017 17:13:41	INFO	HOSTER UploadedTo[3]: Processing url: http://uploaded.net/file/some/somefile.tar
02.03.2017 17:13:41	INFO	HOSTER UploadedTo[3]: Looking for direct download link...
02.03.2017 17:13:41	INFO	HOSTER UploadedTo[3]: Direct download link not found
02.03.2017 17:13:41	ERROR	pycurl error 23: Failed writing body (0 != 1348)
02.03.2017 17:13:41	INFO	ADDON UnSkipOnFail: Looking for skipped duplicates of: somefile.tar (pid:3)
02.03.2017 17:13:41	INFO	ADDON UnSkipOnFail: No duplicates found```

topy:
same here. Issue occuring since a few days

thera314:
same here too with Uptobox.com :(

RickyTR:
Same with easybytez.com premium account.

PedroGPP:
@GammaC0de Is there a solution to this problem?

GammaC0de:
First of all, delete your `userplugins` folder and update to the latest git.

If the problem continues, please provide the below information:
* URL with the issue
* DEBUG log, preferably when run with `python pyLoadCore.py -d`
* Is this a premium download?

thera314:
Hi GammaC0de,

Thanks in advance for your help.

I've just update to the latest git and pb still remains.

URL : http://uptobox.com/6rg96lkwm75m
Premium download : yes

Debug 👍 

03.03.2017 13:32:38	DEBUG	HOSTER UptoboxCom[2]: Plugin status: testing
[...]

PR #2978, opened 2017-10-07
change default octal values to strings

pyblub:
### Type of request:

- [x] Bugfix
- [ ] Code cosmetics
- [ ] Feature enhancement
- [ ] New feature
- [ ] New plugin

### Short description:

change default octal values to strings

replace oct with to_str as values are already strings
int function for base 8 conversion expects a string object

### Additional info:

The values in the default config are integers so using the function oct would work, but when reading in the config file, the values are strings. Function oct expects an integer and throws an error when passing a string.


vuolter:
https://github.com/pyload/pyload/commit/0879f786cc50647b67b635d54d1d86ca50a51c72

GammaC0de:
Shouln't is be:
```
InputType.Octal: lambda x: int(x, 8)
```
This will get a "644" or "0o644" and return 420 decimal, `oct()` returns a string...

pyblub:
At the moment the octal value that is parsed can be an int (from the default config) or a string when parsed from the config file.

To handle both cases you might need a `if isinstance(x, str/int)`. Because the function `int()` cannot handle strings in the format '0oXXX', and `int(x, base=8)` only works with strings. That's why my suggestion was to change them to strings, because in the code the int with a base is used e.g. `int(self.pyload.config.get('permission', 'filemode'), 8)`

@GammaC0de if you do your way, you need a conversion when writing the config as well. Also you have to change the default config octal values to strings, as I did.

It you don't want to use `to_str`, which I don't like either, you can change it to:
`InputType.Octal: lambda x: oct(int(x, 8))`
and change the default octal values to strings by adding quotes, if you want to have some kind of check to see if `x` is actually an octal value.

GammaC0de:
One note: `isinstance()` does not work for base types like int, you should use `if type(X) == int`.

Actually int() does accept strings like "0o644", at  least python 2.7:
```python
>>> int("0o644", 8)
420
```

pyblub:
@GammaC0de `int("0o644", 8)` works and expects a string, but `int("0o644")` does not.
About isinstance (tested with python 2.7 and 3.6):
```
>>> isinstance('8', int)
False
>>> isinstance(8, int)
True
```

GammaC0de:
Well, I had issues with it in the past:
Issue #2766: https://github.com/pyload/pyload/commit/0d342902653c977115583e6e0f35faa7c527ea57

vuolter:
https://github.com/pyload/pyload/commit/1a0babcbdb2cfefae77f24ff84a471a9509f2da6

pyblub:
> 1a0babc

only the changes in  pyload/config/default.py and pyload/config/parser.py are required.

the other changes break everything again.

GammaC0de:
The solution is so simple:
Every type should have its own *set* **and** *get* methods.

pyblub:
@GammaC0de definitely the cleanest solution. Also a validation if the values are correct, would be helpful.

PR #2984, opened 2017-10-13
remove DatabaseBackend object from args

pyblub:
### Type of request:

- [x] Bugfix
- [ ] Code cosmetics
- [ ] Feature enhancement
- [ ] New feature
- [ ] New plugin

### Short description:

remove DatabaseBackend object from args as it is added later on anyway (see. [queue](https://github.com/pyload/pyload/blob/master/pyload/core/database/backend.py#L431) and [async](https://github.com/pyload/pyload/blob/master/pyload/core/database/backend.py#L423))

This problem prevents calling DatabaseBackend.commit, DatabaseBackend.rollback and DatabaseBackend.sync_save.

At the moment it seems that commit is only called when pyload exits. If pyload stops working unexpectedly all information will be lost. Maybe commit after every INSERT/UPDATE?

### DIFF
--- a/pyload/core/database/backend.py
+++ b/pyload/core/database/backend.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import, unicode_literals
 import io
 import os
 import shutil
+from functools import wraps
 from queue import Queue
 
 from future import standard_library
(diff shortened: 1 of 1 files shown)


vuolter:
@pyblub I don't know much about the db module, because was fully written by @RaNaN.
I already seen some "problematic" logics and plan to fix them one day, but for now I still have to restore the webui and, badly, I'll be too busy for the rest of the month to work on. D:

PR #2991, opened 2017-10-27
change atexit decorator to method

pyblub:
### Type of request:

- [x] Bugfix
- [ ] Code cosmetics
- [ ] Feature enhancement
- [ ] New feature
- [ ] New plugin

### Short description:

the decorator points to the class method. when used during initialisation it points to the method of the object and passes self as argument.

before you might have gotten this error:
Error in atexit._run_exitfuncs:
TypeError: exit() missing 1 required positional argument: 'self'

also removed self.config.close() as ConfigParser does not have a close method

PS: self.exit() at line 277 was unreachable

if exit is not called properly, the database does not commit its changes. Maybe committing changes after inserts of packages etc. should be done within the api, otherwise much information will be lost, if pyload crashes unexpectedly.

### DIFF
--- a/pyload/core/base.py
+++ b/pyload/core/base.py
@@ -96,6 +96,8 @@ class Core(object):
 
         self._init_api()
 
+        atexit.register(self.exit)
+
     def _init_api(self):
         from pyload.api import Api
         self.api = Api(self)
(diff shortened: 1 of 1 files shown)


vuolter:
👍

Issue #3000, opened 2017-11-07
LinkFilter Plugin whitelisting


segelkma:
Hi, some time ago I made the LinkFilter Hookplugin:
https://github.com/pyload/pyload/blob/stable/module/plugins/hooks/LinkFilter.py

I would like to add the functionality to invert the behaviour (whitelisting), so that only those links pass, that are in the list. I implemented it but unfortunately it effects all links (crypters and containers). This does not make much sense, so I would like it to work only on hoster plugins. However the Hook API only provides the links in a list and a pid for the event linksAdded:
https://github.com/pyload/pyload/wiki/module.HookManager.HookManager

How do I find out if the link belongs to a hoster plugin? I know that every hoster plugin has a pattern. This is the way that pyload determines which plugin to use for which link right?
But I have no idea on how check them.
Is it even possible within a hook plugin?

Thanks.

GammaC0de:
Untested:
```python
        hosters_dict = self.pyload.pluginManager.hosterPlugins
        links_dict = self.pyload.api.checkURLs(links)
        if any(True for plugin_name in links_dict if plugin_name in hosters_dict):
            pass
```

segelkma:
Thanks, first two lines are working (didn't check third yet). Now I got something to work with.
Can be closed.

segelkma:
Or wait a second:
What would you prefer as an option for the filter list? The current version requests url's to be filtered (e.g. share-online.biz). Would it possibly make more sense to ask the user to input the pluginname into the filterlist (e.g. ShareonlineBiz)?
I can implement either option

GammaC0de:
I think using share-online.biz is more diverse because you can limit just one host that the hoster support.

PR #3004, opened 2017-11-10
Updated Linkfilter plugin (Added whitelist ...)

segelkma:
### Type of request:

- [ ] Bugfix
- [ ] Code cosmetics
- [ ] Feature enhancement
- [x] New feature
- [ ] New plugin
- [x] Plugin enhancement

### Short description:

- Changed default behaviour: Only filters hoster links on default instead of all links
- Added whitelist filter function
- Added option to force execution on all links (previous default behaviour)

https://github.com/pyload/pyload/issues/3000

Tested it successfully with several different hoster links and containers/crypters.
Since the default behaviour changed, does it make sense to rename the plugin? E.g. HosterLinkFilter, HosterFilter, FilterHosterLinks ...

### DIFF
--- a/module/plugins/hooks/LinkFilter.py
+++ b/module/plugins/hooks/LinkFilter.py
@@ -1,13 +1,12 @@
 
 from ..internal.Addon import Addon
 
-import re
 from itertools import chain
 
 class LinkFilter(Addon):
     __name__ = "LinkFilter"
     __type__ = "hook"
-    __version__ = "0.15"
+    __version__ = "0.14"
     __status__ = "testing"
 
     __config__ = [("activated", "bool", "Activated", False),
@@ -40,11 +39,11 @@ class LinkFilter(Addon):
         linkcount -= len(links)
 
         if linkcount > 0:
-            linkstring = '' if self.config.get('forceExecute') else 'hoster '
-            linkstring += 'links' if linkcount == 1 else 'links'
-            self.log_info(
-                'Whitelist filter removed %s %s not containing (%s)' % 
-                (linkcount, linkstring, ', '.join(filters)))
+                linkstring = '' if self.config.get('forceExecute') else 'hoster '
+                linkstring += 'links' if linkcount > 1 else 'link'
+                self.log_info(
+                    'Whitelist filter removed %s %s not containing (%s)' %
+                    (linkcount, linkstring, ', '.join(filters)))
 
     def blacklist(self, links, filters):
         for filter in filters:
@@ -56,7 +55,7 @@ class LinkFilter(Addon):
 
             if linkcount > 0:
                 linkstring = '' if self.config.get('forceExecute') else 'hoster '
-                linkstring += 'link' if linkcount == 1 else 'links'
+                linkstring += 'links' if linkcount > 1 else 'link'
                 self.log_info(
                     'Blacklist filter removed %s %s containing %s' %
                     (linkcount, linkstring, filter))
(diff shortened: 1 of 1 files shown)


module/plugins/hooks/LinkFilter.py
```diff
@@ -22,18 +25,46 @@ def deactivate(self):
...
+            self.blacklist(links, filters)
 
-        for filter in filter_entries:
-            if filter == "":
-                break
+    def whitelist(self, links, filters):
+        linkcount = len(links)
+        links[:] = [link for link in links if
+            not self.isHosterLink(link) or
+            [True for filter in filters if link.find(filter) != -1]]
```
GammaC0de: Shouldn't that be all() or any()?

segelkma: I don't know if it should. Did some testing and it looks like
`[True for filter in filters if link.find(filter) != -1]`
is doing the same as
`any(True for filter in filters if link.find(filter) != -1)`

GammaC0de: This is not true:
```
>>> bool([True, False])
True
>>> bool([False, False])
True
>>> bool([True, True])
True
```
What it actually does is to check if the array is not empty which is not.

So:
`any(link.find(filter) != -1 for filter in filters)`
is  correct and also more readable..

segelkma: What I meant is that I tested both codes using 8 links from different sites (mostly hosters) and changed the plugin settings and filter all the time, then readded the links to queue and compared the outcome. And the filter function haad the same result with both code fragments (Also changed order of links while adding + in filter). However, I agree that the readibilty is better, so in any way it makes sense to change it.

module/plugins/hooks/LinkFilter.py
```diff
@@ -22,18 +25,46 @@ def deactivate(self):
...
                 self.log_info(
-                    "Removed %s %s containing %s" %
+                    'Blacklist filter removed %s %s containing %s' %
                     (linkcount, linkstring, filter))
+
+    def isHosterLink(self, link):
+        #declare all links as hoster links so the filter will work on all links
+        if self.config.get('forceExecute'):
+            return True
+        for name, value in chain(self.pyload.pluginManager.hosterPlugins.iteritems()):
```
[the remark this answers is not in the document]
segelkma: Is this an issue that you want to be resolved?

module/plugins/hooks/LinkFilter.py
GammaC0de: @segelkma Please test this version:
[LinkFilter.zip](https://github.com/pyload/pyload/files/1465255/LinkFilter.zip)

segelkma:
Will do when I get the time, on the weekend maybe.

segelkma:
Some thoughts about your version:
1. Maybe change "Allow only links" to "Allow only links that are...". I think its better for the understanding.
2. Having a different (the old style) in the blacklist function compared to the whitelist one was intentional to have a better logging feature (see which hoster caused which removal). This is by nature not possible for a whitelist filter, so I only coded the whitelist filter as it was. Do you prefer having less logging in blackfilter?
3. In the is_hoster_link function
Do you wish to have this code...
```
for name, value in self.pyload.pluginManager.hosterPlugins.iteritems():
            if value['re'].match(link):
                return True
```
which causes an issue with codacy (unused name variable) but has the better readability or this one...

```
for hoster in chain(self.pyload.pluginManager.hosterPlugins.iteritems()):
            if hoster[1]['re'].match(link):
                return True
```
whichs readability is a little worse but does not cause the issue.

GammaC0de:
👍

Issue #3073, opened 2018-01-23
nginx reverse proxy


kugan49:
I saw some person who have set pyload and nginx to address website via http://servername/pyload/

How can I do this ?

GammaC0de:
Does is have to be nginx  or apache is acceptable too?

kugan49:
I already use nginx (I'm on raspberry Pi) so I can't use anything else. Any ideas ?

GammaC0de:
Ok, so how about this setup?
```
location  /pyload {
  rewrite /pyload(.*) /$1  break;
  proxy_pass         http://localhost:8000;
  proxy_redirect     off;
  proxy_set_header   Host $host;
}
```

kugan49:
when I go on http://myservername/pyload
It redirect to http://myservername/login

kugan49:
I put "/pyload" in path prefix

GammaC0de:
What if you go to http://myservername/pyload/login ?

kugan49:
GOt this

Error 404: Not Found
Sorry, the requested URL 'http://osmc/login' caused an error:

Not found: '//login'

GammaC0de:
Try this:
```
location  /pyload {
  rewrite /pyload(.*) $1  break;
  proxy_pass         http://localhost:8000;
  proxy_redirect     off;
  proxy_set_header   Host $host;
}
```

kugan49:
http://myservername/ redirects to http://myservername/login
and  
http://myservername/pyload/login works but with no css

GammaC0de:
This?
```
location  /pyload {
  proxy_pass         http://localhost:8000;
  proxy_redirect     off;
  proxy_set_header   Host $host;
}
```

kugan49:
This is the same result,

I saw this [https://github.com/pyload/pyload/issues/2631](https://github.com/pyload/pyload/issues/2631) 
But nginx put me an error on the last config (location /pyload/ {
    proxy_pass http://localhost:8000/;
    proxy_set_header Accept-Encoding "";
    sub_filter_types text/css text/xml text/javascript;
    sub_filter '/media/' '/pyload/media/';
    sub_filter '/json/' '/pyload/json/';
    sub_filter '/api/' '/pyload/api/';
    sub_filter '<a href="/' '<a href="/pyload/';
    sub_filter_once off;
}) =

nginx: [emerg] "sub_filter" directive is duplicate in /etc/nginx/sites-available/locat.conf:45

[...]

PR #3077, opened 2018-01-26
Fix path prefix for pyplex/classic UI (#3073)

OzzieIsaacs:
### Type of request:

> **NOTE:**

- [x] Bugfix
- [ ] Code cosmetics
- [x] Feature enhancement
- [ ] New feature
- [ ] New plugin

### Short description:

Path prefix for pyplex UI working, therefore Reverse proxy installations working without huge rewrite rules

as promised

Issue #3073

There is one thing I'm not 100% sure about. Jinjia isn't interpreting css files, so I removed the background-image url to the base.html file, but I haven't found a location where it is used...
After merging please make sure:
- Delete jinjia cache (while pyload is stopped)
- delete website cache of your browser
(99% of all issues with the pathprefix are caused by mixed/chached templates and scripts)


GammaC0de:
👍

PR #3083, opened 2018-02-01
[Stable] Remove required libraries from module/lib

adamantike:
### Type of request:

- [ ] Bugfix
- [x] Code cosmetics
- [ ] Feature enhancement
- [ ] New feature
- [ ] New plugin

### Short description:

Removing the following required libraries from `module/lib`, as they are
already required by `pavement.py`:

* BeautifulSoup
* beaker
* jinja2
* simplejson
* thrift


GammaC0de:
That is true but I don't think it's a wise thing to do because:
- Many users just clone the latest stable branch for installing pyLoad, removing the libraries will cause creation of **lots** of github issues dealing with installation which is already not trivial.
- Some of the latest versions of the libraries are not backward compatible with the versions pyLoad uses, so until pyLoad cannot be installed with `pip install` command this is not a wise thing to do.

Sorry, I have to reject this. 😉

PR #3088, opened 2018-02-02
[Stable] Initial setup.py config

adamantike:
### Type of request:

- [ ] Bugfix
- [ ] Code cosmetics
- [x] Feature enhancement
- [ ] New feature
- [ ] New plugin

### Short description:
    
Adding an initial `setup.py` and `setup.cfg` configuration, to be able to automate installation, packaging and distribution.

A good packaging and distribution solution is required if we want to publish to PyPI, and improve the installation process for end users. It will also simplify the inclusion of other libraries as dependencies.


GammaC0de:
I am not familiar with this subject so if you don't mind, I prefer that @vuolter will review this.

Also, are you sure that BeautifulSoup, Beaker,  jinja2 and thrift should be in the dependencies? because they are already in pyLoad's the lib directory..

adamantike:
@GammaC0de this is mostly based in the [setup.py](https://github.com/pyload/pyload/blob/master/setup.py#L115-L172) file from `master`, and the [pavement.py](https://github.com/pyload/pyload/blob/stable/pavement.py#L20-L63) file from `stable` branch, which specifies the dependencies I added here.

The main goal is to completely remove the `module/lib` directory at some point, as it doesn't scale well if we want to keep adding new dependencies that could simplify some functionalities (database handling, CLI abstraction, test libraries, etc) without copying those packages' sources into this repo.

This is in no way a definitive version for these files, but is a step forward to simplify installation of a development environment by using `setup.py`.

pyblub:
@GammaC0de @vuolter is not really available. He is blocking the master branch and I highly doubt he really knows what he is doing. If possible, you should take over the project completely.

About the dependencies: It is never a good idea to copy the code of other modules directly in a project instead of specifying them as dependencies. It not only helps keeping your code base smaller, but the modules will stay up to date. You also reduce the time spent on maintaining the dependencies.

The only problem I see with changing certain structures in pyload (stable) is, that it might lead to new bugs and the branch becoming unstable. I am not sure what the best approach is to handle this.

That being said, I'm in favour of removing the dependencies and adding an ORM etc.

adamantike:
I can understand the `stable` branch needs to be stable for people to clone the project and run it. This change doesn't affect the execution at all, but other PRs will, and with the `master` branch blocked, there are two solutions I can think of:

- Release and tag commits more often: No need for people to clone the repo, if we release pyLoad versions often. They don't need to follow semver standards at first, if your concerns is not to bump to 0.5, 0.6, etc. Even by publishing 0.4.x versions would allow users to download tagged versions instead of current branch, and allow us to keep moving.
- Create a `develop` branch, and merge as often as we check we don't break anything. That's what `master` should be being used for.

GammaC0de:
@pyblub I had a view words with @vuolter couple of days a go, He is really really busy at work:
>**GammaC0de** Jan 27 22:30
@vuolter Where are you?

>**Walter Purcaro** Jan 30 23:55
 hello, I'm really really sorry I did not show up lately (I mean, in the latest month and over), but I was literally overwhelmed by work and other stuff, I've never been home in practice. O.O

> **Walter Purcaro** Jan 31 00:07
Alright, now you can post on twitter as pyload.

> GammaC0de Jan 31 00:12
:thumbsup: Thanks, nice to see you're back again :grinning:
there is a problem with codacy being stuck again..

> **Walter Purcaro** Jan 31 00:12
Sorry if this took me so long... Currently all the (my) data related to the project are on that damaged drive that I told you time ago and I can do nothing to recover them for now. I just reset the twitter password now. :worried:
.
I'll try to take a look (4 codacy)

As of taking control of the project:
- First of all, I have a lot of respect to @vuolter, I do not want to offend him.
- Secondly, I do not have full permissions to all the things in pyload (codacy for example).
- Thirdly, I am not a professional python coder (in fact I did not knew python at all before I started contributing to pyload) so there are many areas that @vuolter is much better than me - creating an installation release is one of them.

I hope this clarify the situation.

adamantike:
@GammaC0de That doesn't solve or helps in any way to find a solution to the current situation, which is people trying to help being completely blocked, and losing all interest in trying to contribute. The main issue is that development is stalled, trying to help by improving the `stable` branch a lot of times mean duplicating efforts that have already been done in `master`, or waiting for @vuolter for months. If he doesn't have the time, and you do, then there's no reason other than selfishness to open it up to you to keep moving and improving it.

You don't even need to be a **professional Python coder** to be in charge (and not the only one in charge, @vuolter can still be an admin for when he gets the time to come back). You only need to care about the project, to define what it needs, prioritize features and bugs, and be a facilitator for developers to contribute to it. And, as far as I have seen, you are excellent doing that.

I think we are not asking too much, we only want simple ways to contribute to the project without being blocked, making this piece of software better, and provide an excellent solution to its users.

snickers2k:
lol.. i've been waiting for this since years. There is no chance to get some users on board, without even having an automated-installation. sorry, jd2 is bad.. but this situation is worse.

Issue #3122, opened 2018-03-03
API for ReCaptcha v2


funta23:
Here is an answer from 9kw staff to my question if they can help fix 9kw pyload plugin so that hosters using ReCaptcha can work again. 
Not sure what is needed to provide the that API and if work from other mentioned projects might help here.

From 9kw:
> Pyload has no api for recaptcha v2. This problem is from pyload.
> Our captcha service has rcv2 support since months.
> 
> We can only update the plugin for 9kw after the pyload api has rc2 support for captcha services.
> 
> Other tools like jdownloader 2 (since months!) and fritzload (few days) already "fixed".

Michi-F:
I've just looked at their API, it seems to be quite simple:
https://www.9kw.eu/hilfe_more.html

The best news for pyload, running in low power devices:
> All steps have now been completed successfully without browser emulation.
Sounds promising! You don't even need to load a captcha within pyload, you only need the site-key of  the site and do some requests to 9kw.

I'll have a look on how this can be implemented within pyload and the hoster-plugins.

GammaC0de:
@funta23 Can you ask the support why I'm not getting any answer from 9kw (I try every 3 second for a duration of 300 seconds)?
```
session_id = self.load("https://www.9kw.eu/index.cgi?apikey=<9KW_API_KEY>&action=usercaptchaupload&interactive=1&file-upload-01=6LdatrsSAAAAAHZrB70txiV5p-8Iv8BtVxlTtjKX&oldsource=recaptchav2&pageurl=http://www.share-online.biz/dl/WICYFWWOSP")

result = ""
while (!result):
    result  = self.load("https://www.9kw.eu/index.cgi?apikey=<9KW_API_KEY>&action=usercaptchacorrectdata&id=<session_id>")
    time.sleep(3)

```

Michi-F:
Good news: I managed to get 9kw working with share-online: [https://github.com/Michi-F/pyload/tree/stable](https://github.com/Michi-F/pyload/tree/stable)
The code (up to commit [https://github.com/Michi-F/pyload/commit/2ab8644ef8f24233d0b4438c7b6612fd482a7726](Michi-F/pyload @ 2ab8644) is working so far.

So far you can add a new download link in pyload, and everything else is handled by pyload and the 9kw plugin!
Notice that I had to patch the 9kw-plugin, so you have to use my "Captcha_9kW.py" plugin, and not the current one of pyload or from the 9kw-website.

[...]

Issue #3120, opened 2018-03-01
FilerNet: ReCaptcha free downloads not working


funta23:
Seems[ Filer.net switched to ReCaptcha](http://filer.net/changelog) and plugin is not working anymore since some days.
Can anyone help please? 
Already updated to latest version using new [Updatemanager.py ](https://raw.githubusercontent.com/pyload/pyload/353c6692afdec4b6108555725cb5db46d02ed816/module/plugins/hooks/UpdateManager.py)

Sample URL: http://filer.net/folder/ai3b00umflydtexn

```
01.03.2018 22:05:18 INFO      ADDON UpdateManager: pyLoad is up to date!
01.03.2018 22:05:19 INFO      ADDON XFileSharing: Auto-discover new crypters
01.03.2018 22:05:19 INFO      pyLoad is up and running
01.03.2018 22:05:20 INFO      ADDON UpdateManager: All plugins are up to date!
01.03.2018 22:05:21 INFO      Download starts: abcdefghijklmnop.part10.rar
01.03.2018 22:05:21 WARNING   HOSTER FilerNet[2570]: Plugin may be unstable
01.03.2018 22:05:21 INFO      HOSTER FilerNet[2570]: Grabbing link info...
01.03.2018 22:05:22 INFO      HOSTER FilerNet[2570]: Link name: abcdefghijklmnop.part10.rar
01.03.2018 22:05:22 INFO      HOSTER FilerNet[2570]: Link size: 200.00 MiB (209715300 bytes)
01.03.2018 22:05:22 INFO      HOSTER FilerNet[2570]: Link status: online
01.03.2018 22:05:22 INFO      HOSTER FilerNet[2570]: Processing url: http://filer.net/get/abcdefghijk
01.03.2018 22:05:22 INFO      HOSTER FilerNet[2570]: Looking for direct download link...
01.03.2018 22:05:22 INFO      HOSTER FilerNet[2570]: Direct download link not found
01.03.2018 22:05:22 INFO      HOSTER FilerNet[2570]: Checking for link errors...
01.03.2018 22:05:22 INFO      HOSTER FilerNet[2570]: Waiting 1 minutes...
01.03.2018 22:06:24 INFO      HOSTER FilerNet[2570]: No errors found
01.03.2018 22:06:24 INFO      HOSTER FilerNet[2570]: Processing as free download...
01.03.2018 22:06:25 WARNING   CAPTCHA FilerNet[2570]: ReCaptcha | Secure Token pattern not found
[...]

PR #3125, opened 2018-03-07
[API, webinterface] Fix API and webinterface for interactive reCAPTCH…

Michi-F:
### Type of request:

> **NOTE:**

- [ ] Bugfix
- [ ] Code cosmetics
- [x] Feature enhancement
- [x] New feature
- [ ] New plugin

### Short description:

Changes to the API and webinterface to support interactive reCAPTCHA V2
- Add an "interactive" property to the captcha task
- change ReCaptcha plugin
- change webinterface (classic / pyplex)

ReCaptchaV2_userscript.js:
- This is a tampermonkey userscript. To use the feature, the user has to install the add-on tampermonkey in the browser and add this script manually. Updates of the script are handled by the tampermonkey add-on (`@updateURL` setting points to github, so a commit to github will trigger an update for the user automatically).

Handling of interactive reCAPTCHA V2 captchas are not supported by the current API

https://github.com/pyload/pyload/issues/3074
https://github.com/pyload/pyload/issues/3122
https://github.com/pyload/pyload/issues/3120


module/Api.py
```diff
@@ -805,7 +806,11 @@ def getCaptchaTask(self, exclusive=False):
         if task:
             task.setWatingForUser(exclusive=exclusive)
             data, type, result = task.getCaptcha()
-            t = CaptchaTask(int(task.id), standard_b64encode(data), type, result)
+            if task.isInteractive() == 1:
```
pyblub: remove `== 1`. task.isInteractive() returns a boolean value

module/CaptchaManager.py
```diff
@@ -62,7 +62,10 @@ def handleCaptcha(self, task):
         cli = self.core.isClientConnected()
 
         if cli: #client connected -> should solve the captcha
-            task.setWaiting(50) #wait 50 sec for response
+            if task.isInteractive() == 1:
```
pyblub: remove `== 1`. task.isInteractive() returns a boolean value

module/CaptchaManager.py
```diff
@@ -62,7 +62,10 @@ def handleCaptcha(self, task):
...
+                task.setWaiting(120) #wait 120 sec for response, interactive captcha can take some time
```
pyblub: even 120 seconds is sometimes not enough at least for me :)

module/plugins/captcha/ReCaptcha.py
```diff
@@ -354,3 +356,45 @@ def _challenge_v2(self, key, secure_token=None):
...
+    # solve interactive captcha (javascript required), use when non-JS captcha fallback for v2 is not allowed
+    def _challenge_v2_js(self, key):
+        self.log_debug("challenge ReCaptcha v2 interactive")
+        captchaManager = self.pyload.captchaManager
+
+        try:
+            time_ref = ("%.2f" % time.time())[-6:].replace(".", "")
+            filename = "captcha_request_%s_%s.nofile" % (self.pyfile.plugin.__name__, time_ref)
+            # we don't have an actual image to decrypt -> pass the URL and the site-key for further use in plugins/hooks
+            img = {'url': self.pyfile.url, 'sitekey': self.retrieve_key(None)}
```
pyblub: why not use `key` as it has been passed to the method. or remove key from the parameters.

module/plugins/captcha/ReCaptcha.py
```diff
@@ -354,3 +356,45 @@ def _challenge_v2(self, key, secure_token=None):
...
+            self.task = captchaManager.newTask(img, "interactive", filename , "interactive")
```
pyblub: pass `None` instead of filename as it is not used anyway. then you can remove line 366 and 367

module/plugins/captcha/ReCaptcha.py
```diff
@@ -354,3 +356,45 @@ def _challenge_v2(self, key, secure_token=None):
...
+
+            captchaManager.handleCaptcha(self.task)
+
+            # @TODO: Move to `CaptchaManager` in 0.4.10
+            self.task.setWaiting(120)
```
pyblub: As before maybe increase 120 seconds.

module/plugins/captcha/ReCaptcha.py
```diff
@@ -354,3 +356,45 @@ def _challenge_v2(self, key, secure_token=None):
...
+            while self.task.isWaiting():
+                self.pyfile.plugin.check_status()
+                time.sleep(1)
+
+        finally:
+            captchaManager.removeTask(self.task)
+
+        result = self.task.result
+
+        if self.task.error:
```
pyblub: This code is copied from Captcha.py. Can you instead create a method `check_captcha_result` or something in Captcha.py and then use that method in Captcha.decrypt_image and here, so that we do not duplicate the code.

module/web/media/js/classic/recaptcha-v2-interactive.js
```diff
@@ -0,0 +1,90 @@
+﻿/*
```
pyblub: This file contains tabs instead of spaces and is identical to module/web/media/js/pyplex/recaptcha-v2-interactive.js
Can you put it in the parent directory and reuse it in both themes?

module/web/media/js/pyplex/recaptcha-v2-interactive.js
```diff
@@ -0,0 +1,90 @@
+﻿/*
```
pyblub: This file contains tabs instead of spaces and is identical to module/web/media/js/classic/recaptcha-v2-interactive.js
Can you put it in the parent directory and reuse it in both themes?

module/web/media/pyplex/css/styles.css
```diff
@@ -451,4 +457,20 @@ textarea.form-control:focus {
 .btn[disabled],
 fieldset[disabled] .btn {
   pointer-events: auto;
+}
+
+iframe#interactiveCaptchaIframeId
+{
+  position:relative;
+  width:465px;
```
pyblub: Do we need hardcoded values?

module/Api.py
```diff
@@ -805,7 +806,11 @@ def getCaptchaTask(self, exclusive=False):
...
+                data = json.dumps(data) #data = {url:..., sitekey:...}
```
pyblub: inline comments in python should have at least to spaces before the hash sign and one afterwards for readability.
e.g. `data = json.dumps(data)  # data = {url:..., sitekey:...}`

module/plugins/captcha/ReCaptcha.py
```diff
@@ -354,3 +356,45 @@ def _challenge_v2(self, key, secure_token=None):
             self.fail(_("Recaptcha max retries exceeded"))
 
         return result, challenge
+
+    # solve interactive captcha (javascript required), use when non-JS captcha fallback for v2 is not allowed
+    def _challenge_v2_js(self, key):
```
pyblub: remove `key` from the parameters as it is not used, because the key is retrieved later on.

module/plugins/captcha/ReCaptchaV2_userscript.js
```diff
@@ -0,0 +1,99 @@
...
+                return grecaptcha && grecaptcha.getResponse().length !== 0;
+            };
+
+            window.captchaOnloadCallback = function()
+            {
+                grecaptcha.render(
+                    document.getElementById("captchadiv"),
+                    {sitekey:window.pyloadRecaptchaV2InteractiveSitekey, callback: pyloadCaptchaCompletedCallback}
+                );
+
```
pyblub: Line 87 contains whitespaces

module/plugins/captcha/ReCaptcha.py
Michi-F: Thought the same, but I saw that e.g. the 9kw-Plugin uses the image-name to detect the hoster of the CaptchaTask.
"captcha_request_ShareOnlineBiz_0123456.jpg"
The plugin extracts "ShareOnlineBiz" to apply hoster-dependet settings of 9kw.

I don't know if any other plugins also rely on the informations "encoded" in the image-name

If there is a other way to get the hoster in the 9kw-plugin, then the image name can be set to None. Maybe this requires to add a "hoster"-attribute to the CaptchaTask-Class in CaptchaManager.py.
Passing this "filename" was just an easy workaround arount this ....

module/web/media/js/classic/recaptcha-v2-interactive.js
Michi-F: Damn, I thought I removed all tabs, but missed that one :( sorry.

But the two files are actually not 100% the same.... the one in pyplex uses $ from jquery to access the DOM, the other (classic template) uses $ from mootools (as jquery is not available in the classic template). And the API to acess the elements is different (jquery uses `#` to access an element by ID, mootools does not).

It's the same for the other scripts (like base.js), which are 80% the same code in both templates, but the DOM access is different, and therefore the files are "duplicated".

The file could we rewritten to be independent of jquery and mootools.... then it could be used for both templates. But I think that's a question of where pyload is heading.

If the old/classic theme will be dropped at some point and replaced by pyplex, I think it's better to use the functionality of jquery in pyplex (instead of writing everything in plain javascript), and make the necessary adjustments for the classic template in a seperate file.

If not, we can think about writing the script in plain javascript.
But that means limiting the possibilities of the more modern pyplex for the support of the old classic template....

I don't know whats the best for pyload here, maybe that's a decision which more people should be involved.

module/web/media/pyplex/css/styles.css
Michi-F: Don't know, something that should be tested. 
I measured the size of the captcha on my screen, and limited the iframe to that size.

Don't know if the size of google captcha is the same on all devices. E.g. on high-resolution displays or tablets it may be different.

I agree that this should be changed, but  I don't know a quick solution of how to get the size of the captcha (before the checkbox to expand it).

module/plugins/captcha/ReCaptcha.py
Michi-F: Maybe the whole code of the _challenge_v2_js function should be moved to Captcha.py to a decrypt_interactive (or something like this) function, to share the same code with decrypt_image there. And in ReCaptcha.py we need only to call the new function.

That's a design decision.... maybe more people can give their opinion on this.

pyblub: Captcha.py is the base class and should only provide general functionality to handle captchas, while ReCaptcha.py is more specific. Captcha.py should not contain anything that is specific to ReCaptcha only, like "sitekey".

GammaC0de: @pyblub Look at the changes I pushed to the 'all ' branch.
ReCaptcha will first try the old `fallback` method and if fails will try `2js` version (I renamed v2_javascript to 2js).
The advantages of this are:
It is more general version, no need to change the hoster to explicitly use 2js.
It can save 9kw credits because it will use 9kw's recaptchav2 only when necessary.

PR #3127, opened 2018-03-07
[Hook] Fix 9kw.eu plugin, update to latest version from 9kw.eu and ad…

Michi-F:
…d support for interactive reCAPTCHA V2

### Type of request:

> **NOTE:**

- [ ] Bugfix
- [ ] Code cosmetics
- [ ] Feature enhancement
- [ ] New feature
- [ ] New plugin

### Short description:

Update 9kw.eu plugin, update to latest version from 9kw.eu and add support for interactive reCAPTCHA V2 https://github.com/pyload/pyload/pull/3125

Handling of interactive reCAPTCHA V2 captchas are not supported by the current plugin

https://github.com/pyload/pyload/issues/3074
https://github.com/pyload/pyload/issues/3122
https://github.com/pyload/pyload/issues/3120

### DIFF
--- a/module/plugins/hooks/Captcha9Kw.py
+++ b/module/plugins/hooks/Captcha9Kw.py
@@ -1,254 +1,309 @@
 # -*- coding: utf-8 -*-
 
+"""
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 3 of the License,
+    or (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+    See the GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program; if not, see <http://www.gnu.org/licenses/>.
+
+    @author: mkaay, RaNaN, zoidberg
+"""
 from __future__ import with_statement
 
-import base64
-import re
+from thread import start_new_thread
+from base64 import b64encode
 import time
 
+try:
+    from urllib.parse import urlsplit
+except ImportError:
+    from urlparse import urlsplit
+
+from module.network.RequestFactory import getURL
 from module.network.HTTPRequest import BadHeader
 
-from ..internal.Addon import Addon
-from ..internal.misc import threaded
-
+from module.plugins.internal.Addon import Addon
+from module.plugins.internal.misc import threaded
+import re
 
 class Captcha9Kw(Addon):
     __name__ = "Captcha9Kw"
     __type__ = "hook"
-    __version__ = "0.38"
-    __status__ = "testing"
-
-    __config__ = [("activated", "bool", "Activated", False),
-                  ("check_client", "bool", "Don't use if client is connected", True),
+    __version__ = "0.44"
+    __description__ = """Send captchas to 9kw.eu"""
+    __config__ = [("activated", "bool", "Activated", True),
+                  ("force", "bool", "Force captcha resolving even if client is connected", True),
+                  ("https", "bool", "Use HTTPS (secure connection)", False),
                   ("confirm", "bool", "Confirm Captcha (cost +6 credits)", False),
                   ("captchaperhour", "int", "Captcha per hour", "9999"),
                   ("captchapermin", "int", "Captcha per minute", "9999"),
-                  ("prio", "int", "Priority (max 10)(cost +0 -> +10 credits)", "0"),
-                  ("queue", "int", "Max. Queue (max 999)", "50"),
-                  ("hoster_options",
-                   "str",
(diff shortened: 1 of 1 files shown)


module/plugins/hooks/Captcha9Kw.py
```diff
@@ -1,254 +1,309 @@
...
         else:
-            self.log_error(res)
+            self.log_error(response)
             return 0
 
     @threaded
-    def _process_captcha(self, task):
-        try:
+    def processCaptcha(self, task):
+        result = None
```
pyblub: variable results is not used and can be removed

module/plugins/hooks/Captcha9Kw.py
```diff
@@ -1,254 +1,309 @@
...
+        if task.isPositional():
+            mouse = 1
+        else:
+            mouse = 0
+
+        regex = re.compile("_([^_]*)_\d+.\w+")
+        r = regex.search(task.captchaFile)
+        pluginname = r.group(1)
+
+        min_option = 1
```
pyblub: The following code is rather messy.
I think you might have used `hosteroptions` instead of `hosteroption` in some `elif` blocks.

In my opinion this should be changed in a way that a dictionary is used, as a dictionary is passed to `getURL` with the parameter `post` anyway. The values within the dictionary can be updated, depending on the values set within `hosteroptions_conf`.

The innermost if statement uses `hosteroption[1].isdigit()` for every condition. this could be checked one time before the if, by using `continue` if `hosteroption[1].isdigit()` is false.

`if len(hosteroptions) > 0:` is not necessary, because the for loop would not do anything anyway if `len` is zero.

module/plugins/hooks/Captcha9Kw.py
```diff
@@ -1,254 +1,309 @@
...
+                    "pageurl" : pageurl,
+                    "interactive" : interactive,
+                    "action": "usercaptchaupload" })
+                time.sleep(2)
+
+                if(response != ""):
+                    break;
 
             except BadHeader, e:
                 time.sleep(3)
```
pyblub: set response to None, otherwise it will not be set at all in case of an exception

module/plugins/hooks/Captcha9Kw.py
```diff
@@ -1,254 +1,309 @@
...
 
-            else:
-                if res and res.isdigit():
-                    break
+        if response.isdigit():
```
pyblub: Check if `response is not None`

module/plugins/hooks/Captcha9Kw.py
```diff
@@ -1,254 +1,309 @@
...
-            if not result or result == "NO DATA":
-                time.sleep(5)
-            else:
-                break
+                if response2 == "" or response2 == "NO DATA":
+                    time.sleep(5)
+                else:
+                    break;
 
+            result = response2
```
pyblub: this seems unnecessary. Just continue using response2 instead of result

module/plugins/hooks/Captcha9Kw.py
```diff
@@ -1,254 +1,309 @@
...
 
+            if self.config.get("timeout") > 0:
+                timeout_option = self.config.get("timeout")
+            else:
+                timeout_option = 300
+
+            hosteroptions_conf = self.config.get("hosteroptions")
+            hosteroptions = hosteroptions_conf.split(";")
+
+            if len(hosteroptions) > 0:
```
pyblub: similar to before. this statement is superfluous due to the for loop afterwards.

module/plugins/hooks/Captcha9Kw.py
```diff
@@ -1,254 +1,309 @@
...
+                for hosterthing in hosteroptions:
+                    hosterdetails = hosterthing.split(":");
+                    if len(hosterdetails) > 0 and hosterdetails[0].lower() == pluginname.lower():
+                        for hosterdetail in hosterdetails:
+                            hosteroption = hosterdetail.split("=");
+                            if len(hosteroption) > 1 and hosteroption[0].lower() == 'timeout' and hosteroption[1].isdigit():
+                                timeout_option = hosteroption[1]
```
pyblub: only timeout option is used here, therefore you could add a `break` statement

module/plugins/hooks/Captcha9Kw.py
```diff
@@ -1,254 +1,309 @@
...
-        if not self.config.get('passkey'):
-            return
+        if self.getCredits() > 0:
+            if self.config.get("queue") > 10 and self.config.get("queue") < 1000:
+                https = "https" if self.config.get("https") else "http"
+                servercheck = getURL(https + "://www.9kw.eu/grafik/servercheck.txt", get = {})
 
-        if self.pyload.isClientConnected() and self.config.get('check_client'):
-            return
+                for i in range(1, 3, 1):
```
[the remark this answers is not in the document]
pyblub: Due to @vuolter adding these hooks, him not being availble and him being the only person with all permissions on this repo, it's not possible to merge unless all webhooks report back that the change can be merged. One of the reasons why I'm in favour of forking this repo.

PR #3174, opened 2018-05-28
Realdebrid

synopat:
### Type of request:

- [ ] Bugfix
- [ ] Code cosmetics
- [ ] Feature enhancement
- [ ] New feature
- [ ] New plugin

### Short description:

1. Verify free-account correctly.
2. Able to login with username and password (instead of given api-token as password).

### DIFF
--- a/module/plugins/accounts/RealdebridCom.py
+++ b/module/plugins/accounts/RealdebridCom.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 
 import time
+import re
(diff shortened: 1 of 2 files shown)


audren49:
Hi.

It seems this request and the related commits have removed the ability to login with a premium account.
Using the API token now result in an unvalid account, not premium.
Using the account password result in an valid account, not premium.

Using the API token used to result in a valid account, premium.

Cannot use real debrid premium anymore as a consequence.

PR #3244, opened 2018-10-17
[Fix] Correct SERVER_URL to plugins.txt to avoid http redirect

revog:
Changed the SERVER_URL to raw URL of file plugins.txt to avoid http redirect and failing update process (leds to exception)


GammaC0de:
Please increase plugin version.

GammaC0de:
👍

PR #3268, opened 2018-11-12
Shipped libraries don't need to be installed

lukastribus:
jinja2, beaker and thrift are shipped with pyLoad. Make that clear.


GammaC0de:
Thanks :+1:

PR #3316, opened 2019-01-19
Update README.MD

Velociraptor85:
since 2.7 is the minimum as far as I know

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -14,7 +14,7 @@ Targeted for headless NASs it is designed to be extremely lightweight, fully cus
 
 # Dependencies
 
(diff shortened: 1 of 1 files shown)


GammaC0de:
There a still few users on old hardware that keep using python 2.5.

As far as I know, pyload can still run on python 2.5 (there is one exception to keep in mind: if you use the `with` statement in your code then you should do `from __future__ import with_statement`)

So  the correct answer is 2.5 may run with som TLC.

Velociraptor85:
OK great =) I just thought because I read some time ago that 2.5 would be dropped ^^

PR #3338, opened 2019-03-12
[IRC/XMPP] Module/plugins/hooks/irc xmpp

slmi:
Hello,

XMPP update version 0.22
IRC update version 0.27

Modify command :
- packinfo by name or id
- add by name or id
- config maxline for command more
- rename start/stop in unpause/pause

Add command:
- getLog, restartPackage by name or id, restartFile, deleteFinished, restart, freeSpace, togglepause

Add function: 
- exit

Fix:
- del -l

### DIFF
--- a/module/plugins/hooks/IRC.py
+++ b/module/plugins/hooks/IRC.py
@@ -31,7 +31,8 @@ class IRC(Thread, Notifier):
                   ("owner", "str", "Nickname the Client will accept commands from", "Enter your nick here!"),
                   ("info_file", "bool", "Inform about every file finished", False),
                   ("info_pack", "bool", "Inform about every package finished", True),
-                  ("captcha", "bool", "Send captcha requests", True)]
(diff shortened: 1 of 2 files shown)


module/plugins/hooks/IRC.py
```diff
@@ -378,23 +377,119 @@ def event_c(self, args):
         task.setResult(" ".join(args[1:]))
         return ["INFO: Result %s saved." % " ".join(args[1:])]
 
+    def event_freeSpace(self, args):
+        return ["INFO: Free space is %.3f G." % (int(self.pyload.api.freeSpace())/1024/1024/1024)]
```
GammaC0de: Youd'd better import and use `format_size()` from [misc.py](https://github.com/pyload/pyload/blob/stable/module/plugins/internal/misc.py#L273) here.

slmi:
Is already imported, modified accordingly.
Thanks for your time.

PR #3354, opened 2019-03-31
Libgen plugin

yannouchou:
This is a proposal for a Libgen.io download plugin (non-comics).
It will loop across libgen mirrors to attempt the downloads, and process them sequentially to comply with rate limits.


yannouchou:
Since I mixed up my branches I decided to close this PR in favor of #3355

PR #3355, opened 2019-03-31
Libgen (fiction and comics)

yannouchou:
And this one is for Libgen comics - this one has no special logic, just the basic Http plugin with rate limiting.


yannouchou:
All the changes from #3354  are also integrated here so let's have a single PR for LibGen (non-comics and comics).
* The Comics flavor will only apply the rate limiting (1 single download at a time).
* The non-Comics flavor will detect the desired books' MD5 and then loop through known LibGen mirrors to get them, applying Content-Disposition renaming and rate limiting.

yannouchou:
Hello there.
I just pushed a minor update.
Any chance to have this integrated?

GammaC0de:
> Any chance to have this integrated?

Can you please describe what is this site?
Is there interest for others for this plugin?

yannouchou:
libgen.io is arguably the largest source or ebooks (for magazines, fiction, and research papers) and comics.
We can easily download files manually from the website directly but using a download manager for this site is very useful because:

- It is often useful to download large batches of files (e.g. for series, especially with comics), and there's a limit of 1 parallel download on the site so it's inconvenient to do it manually
- It allows to send the downloads to a seedbox / mediacenter where pyload is running, where an ebook manager might be running (in my case I use a Calibre server on my seedbox, where I also have Plex/Sonarr/Radarr for other media).
- It allows bypassing the ad screen of the site.
- It allows looping through different mirrors of the site, as some may be down occasionally.

I've been using it intensively now for a few weeks and it works really well - congrats btw for Pyload, it's working really well for my purposes!

module/plugins/hoster/LibGenComics.py
```diff
@@ -0,0 +1,45 @@
...
+    __status__ = "testing"
+
+    __pattern__ = r'https?://libgen.io/comics0/.+\.(cbr|cbz|zip|rar|pdf)'
+    __config__ = [("activated", "bool", "Activated", True)]
+
+    __description__ = """Download comics from libgen.io respecting throttling limits"""
+    __license__ = "GPLv3"
+    __authors__ = [("Yann Jouanique", "yann.jouanique@gmail.com")]
+
+    def setup(self):
```
GammaC0de: Please remove `multi_dl` and `limit_dl`, they are not used.
```
   def setup(self):
        self.chunk_limit = -1
        self.resume_download = True
        self.multiDL = False
        self.limitDL = 1
```

GammaC0de: `multiDL` is `False` so `limitDL` is unnecessary.
```
   def setup(self):
        self.chunk_limit = -1
        self.resume_download = True
        self.multiDL = False
```

GammaC0de:
It would be better if you can combine this into one single plugin and act accordingly based on the `pyfile.url` given.

module/plugins/hoster/LibGenComics.py
yannouchou: OK, I wasn't entirely sure of the usage but indeed looks like `multiDL` is enough.

yannouchou:
I think it would be a bit convoluted to club both in the same class. Even though the site is the same (though actually the real links don't end up on the same hosts), the download mechanisms are completely different:

- books are assigned an md5 hash that is used to look up the file individually and uniquely across different mirrors, and the files stored in a structured db, with an API that could be queried to get book info (I'm not currently using it, but I may in the future)
- comics (actually, just the "unsorted" section of the comics library, which is only under the `/comics0` path) are just a file server with completely unsorted files in a directory structure. There are no mirrors, no md5 hash, but there is possible directory recursion (which I've just implemented) and batch downloads

So there is actually nothing really common in the download logic, with the exception of the name, end the use of `self.multiDL = False`

GammaC0de:
If you don't mind, I would like to try to build a combined plugin (after I finish fixing some other plugins).

yannouchou:
Yes, actually I've thought about this further, and I agree...
I'll try to work on this, probably next week.

GammaC0de:
Thanks for your contribution.

If I understand correctly, `LibGenComics.py`  is now obsolete, right?
If so, you should remove it from the PR before it can be merged..

yannouchou:
Yes indeed. Sorry I sent the commit without much explanations, but I've indeed merged the logic for both comics and non-comics, with many improvements a things are working quite well.

There is however one pending issue that I haven't been able to sort out. This website frequently has double-byte characters in their URLs which are URL-encoded as expected with 2 percent-encoded bytes (e.g. `é` = `%c3%a9`), and I've managed to handle these when downloading **files** using calls to `self.download()` by specifying `fixurl=False`.
However, when I'm making HTTP calls to web pages to scrape them instead of downloading a file, using `self.load()`, the `fixurl` option is not present, and all attempts to fetch pages that have a double-byte URL encoding fail.
So for instance:
* Fetching `http://libgen.io/comics0/_0DAY/French%20Comics/` works fine
* But fetching `http://libgen.io/comics0/_0DAY/_En%20cours%20de%20t%c3%a9l%c3%a9chargement/` fails

I really couldn't figure that out, and I don't really see any reason why this would fail, since plain curl handles these perfectly well. Any ideas?

yannouchou:
Hmm I just saw #2138 which might address my issue...

GammaC0de:
> Hmm I just saw #2138 which might address my issue...

Unicode is support is the one of the oldest and painful bugs that pyload has :(

PR #3401, opened 2019-06-10
Fix LibGen plugin filename

yannouchou:
Just noticed that the filename is wrong (case doesn't match with the class name) in the repo.
Strangely it was correct on my local version, and hence was working fine, but it was wrong in the commits... I guess I had made local renames git didn't push them since I was on Windows and made no difference...


GammaC0de:
If we are at dealing with proper name for the plugin,
the policy for naming plugins is the site URL with capital letter for every word of it.
So for `libgen.io` the file name should be `LibgenIo.py` and the class name would be `LibgenIo`.

To fix this I would suggest that we delete `Libgen.py` from git (delete, not rename, so it would also be deleted from the clients computers by `UpdateManager`) and create a new plugin named `LibgenIo.py` with all the changes above.

What do you say?

yannouchou:
Sure, makes sense, I didn't notice the convention but I'm all for following those.

GammaC0de:
Thank you 👍

PR #3419, opened 2019-07-13
fix issubclass and resource_filename

grahamannett:
this only fixes parts of it, are you taking PR's?  for me you still need to add https://github.com/ahupp/python-magic and libmagic as well as some of the imports, unless this breaks other things?


vuolter:
Hi @grahamannett thank you for your contribution.
Right now I'm not working on pyload, but, I know, I've left several changes pending last time, so I agree, I've to find time to fix it all.

PR #3420, opened 2019-07-15
feature/add-addToPackage-button - Add an "add to package" button

rubenOF:
This pull request adds a new button to the packages on the queue and collector pages. With this button one can add new links to an existing package. Right now only the "modern" theme support this button. Other theme's will be added if I get an OK from one of the pyload developers.

### DIFF
--- a/module/web/json_app.py
+++ b/module/web/json_app.py
@@ -212,6 +212,25 @@ def edit_package():
         return HTTPError()
 
 
+@route("/json/add_to_package", method="POST")
+@login_required('ADD')
+def add_to_package():
+    try:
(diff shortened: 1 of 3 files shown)


GammaC0de:
Changing a package after it have already been created can result in some unexpected behavior (for example while in the middle of download)

Also, why would you want to add links to an existing package?

rubenOF:
It is pretty much a copy of /api/addFiles, so I dont see why this can result in unexpected behavior. 
The reason why I've added this functionality is because I missed it. I'm new to the pyload project, I tested it out and was unable to fix the typo's I had in links. The only way this could be fixed was by removing and re-adding the package again.

PR #3443, opened 2019-09-10
Do not overwrite sys.excepthook when started from pyLoadGui

snilt:


GammaC0de:
What is the purpose of this PR?
Why would you not overwrite sys.excepthook when started from pyLoadGui?

snilt:
When some other gui application, that installs its own exeption hook, imports the server class and then runs the server inside that application, the exeption hook function will be replaced by the one that belongs to the server which is not wanted. Actually that is the case with my fork of the pyload gui.

PR #3459, opened 2019-10-27
Remove leading tick

letitplaylikecrazy:
Fix linksnappy account plugin


GammaC0de:
Oopsi :man_facepalming:

Thanks :) :tulip:

PR #3515, opened 2020-02-17
Fixes for making application start

StefftheEmperor2:
* Modified .gitignore to ignore local files generated by make_locales and my ide settings.
* initialize babel
* fix for type mismatch in utils/seconds

Signed-off-by: Stefan Bischoff <stefan.bischoff@laser-line.de>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -159,6 +159,7 @@ coverage.xml
 
 # Translations
 *.mo
+src/pyload/locale
 #*.pot
 
 # Django stuff:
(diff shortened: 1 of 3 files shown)


vuolter:
I think this PR can be merged directly to master, I'm just not 100% sure about the reason behind the additional rule to ignore the locale files. I suppose we can just add `*.po` instead of the path.

Anyway I prefer to keep the current lists unchanged because those rule blocks are taken directly from https://github.com/github/gitignore. Ofc you can add any new rule you want at the end of the file, especially since i don't see conflicts here. Thank you.

Ah, of course, thanks for your contributions. I took a look at your repo, looks super interesting! 😉

Issue #3582, opened 2020-04-13
Can't add my premium ddl.to account.


Vinnpinn:
### Describe the problem

I'm trying to add my ddl.to premium account to pyload. I've set it up to run in a docker container on my qnap NAS. When I try to add the account to pyload it says not valid in status and no under premium.
I've tried adding the same account in JDownloader and that works.

### Debug log

```
13.04.2020 18:03:38 INFO      ACCOUNT DdlTo: Adding user `vin*******`...
13.04.2020 18:03:38 INFO      ACCOUNT DdlTo: Relogin user `vin*******`...
13.04.2020 18:03:38 ERROR     ACCOUNT DdlTo: Could not login user `vin*******` | Bad server response: 503 Service Unavailable

```

GammaC0de:
Are you using some kind of VPN?

Vinnpinn:
No I don't. I even tried running pyload on my windows machine and I get
the same error. I can connect to ddl.to without any problems and downloads
with the premium account work fine from jdownloader, running form the same
network.

Am Mo., 13. Apr. 2020 um 21:54 Uhr schrieb GammaC0de <
notifications@github.com>:

GammaC0de:
Account information is needed for testing, please set to a temporary password and then send account user and password to nitzo2001 at yahoo dot com.

Can you provide debug logs?

Vinnpinn:
I sent you my account data, attached is the log file after enabling debug mode and trying to add the account.
[log.txt](https://github.com/pyload/pyload/files/4471668/log.txt)

Tockra:
Same issue here. Since a while I can't login with pyload to ddl.to ...

GammaC0de:
I can confirm this is a cloudflare issue.

Tockra:
Can you solute this problem?

GammaC0de:
Now it works - it seems they turned cloudflare off at-the-moment..

Vinnpinn:
Can confirm that it does work now, downloads work aswell.

![image](https://user-images.githubusercontent.com/26883610/79284863-2ea8fe80-7ebc-11ea-9f8b-ce67116ec314.png)

Tockra:
ddl.to still make problems:
```bash
pyload_1  | 15.04.2020 02:15:29 INFO      Download restarted: name.2019.GERMAN.1080P.WEB.H264-WAYNE.part6.rar | Fallback to free processing
[...]

PR #3586, opened 2020-04-16
[plugins] ddl.to new API-Address

d30r:
[pyload/pyload] Can't add my premium ddl.to account. (#3582)


GammaC0de:
Increase version please

PR #3595, opened 2020-04-22
fix TRAFFIC_LEFT_PATTERN for ddownload.com not matching negative numbers

Phist0ne:
Traffic left for ddownload.com may be negative, if the last downloaded file was larger than the traffic left. This will match the negative traffic left instead of falling back to unlimited.

I was also thinking about capping the traffic value to 0 if a negative value is matched, but this has to be done in XFSAccount.py (or by copying the whole grab_info method).. This seems to work as is, as the negative traffic does not match the first pattern in misc.parse_size, which already short circuits to 0.


GammaC0de:
Increase plugin version please

Phist0ne:
I bumped the version. I could not test this on an account with traffic, as my account is stuck at -123 for 5 more hours.

PR #3648, opened 2020-05-23
Check for running process on windows

snilt:
On windows, accidentally starting a second instance of pyload with same the configdir wipes the pidfile.
Code adopted from psutil (psutil_pid_is_running + psutil_check_phandle), some parts omitted:
  https://github.com/giampaolo/psutil/blob/994c429c010049a6a0556ce3b0d1af1f86f27867/psutil/arch/windows/process_utils.c


GammaC0de:
👍

PR #3674, opened 2020-06-23
Thrift fix: eta data

snilt:
To reproduce the issue:

1. Start a server with an empty queue
2. From the webinterface:
Pause the server
Disable download speed limit
Add some valid package to download (preferably some large file, couple of gigabytes)
3. Start the CLI and unpause the server

...
...
  File "/home/snilt/pyload/module/lib/thrift/protocol/TBinaryProtocol.py", line 110, in writeI32
    buff = pack("!i", i32)
error: 'i' format requires -2147483648 <= number <= 2147483647

It seems that the server sends some bad eta value at the very beginning of a new download.
The PR just changes the data format to integer 64-bit.
(Works fine for me since years)

### DIFF
--- a/module/remote/thriftbackend/pyload.thrift
+++ b/module/remote/thriftbackend/pyload.thrift
@@ -68,7 +68,7 @@ struct DownloadInfo {
   1: FileID fid,
   2: string name,
   3: i64 speed,
-  4: i32 eta,
+  4: i64 eta,
   5: string format_eta,
   6: i64 bleft,
   7: i64 size,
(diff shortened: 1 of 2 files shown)


vuolter:
This PR is for pyLoad Stable, but I want to inform you that thrift is no longer supported in pyLoad Next.

GammaC0de:
> but I have to inform you that thrift is no longer supported in pyLoad Next.

@vuolter Thrift is also used by the Android app...

vuolter:
I think we can add it as a separate package, even I actually don't like it so much to be honest.

PR #3717, opened 2020-09-03
[UptoboxCom] multi downloads require premium account

dguglielmi:
Multi Downloads require a premium account, otherwise we start waiting for concurrent downloads but only one really start and others fails.


GammaC0de:
Increase version please

dguglielmi:
> Increase version please

done

PR #3747, opened 2020-09-19
[Main] Fix for nonworking ClickNLoad

OzzieIsaacs:
### Describe the changes

Added `return "true"` to /flash/add routine in case is succeds, changed `self.pyload.config.get("webui", "use_ssl")`, before it was parameter 'https' which isn't a valid parameter

Issue ##3724

### DIFF
--- a/src/pyload/plugins/addons/ClickNLoad.py
+++ b/src/pyload/plugins/addons/ClickNLoad.py
@@ -133,7 +133,7 @@ class ClickNLoad(BaseAddon):
                             socket.AF_INET, socket.SOCK_STREAM
                         )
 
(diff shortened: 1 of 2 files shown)


GammaC0de:
Thanks for noticing and your PR.
According the the `ClickNLoad` specs, the `/add` method should return `success\r\n` on success and `failed` on failure, I allowed myself to add a fixed solution.

Again, thanks a lot for your contribution, it is much appreciated 👍

OzzieIsaacs:
Yes, you know the code much better than I do, so I'm aware that you might find better solutions than I do.

PR #3748, opened 2020-09-19
Patch delete files

OzzieIsaacs:
### Describe the changes

Change some missing api endpoint in default theme. Also changed api calls from `this.id` to `this` to make the function calls to `delete_files`, `restart_files` work

The problem only occurs in Default Theme if you try to delete files or restart files in a package. A request to `delete_files/[undefined]` is issued and subsequently the function fails.
grafik
 -->


GammaC0de:
Shouldn't delete_packages also use this instead of this.id?

OzzieIsaacs:
Delete packages worked, so I didn‘t debug it or checked if this would also work

PR #3749, opened 2020-09-20
[Main] Fix image captcha with hoster nitroflare 

OzzieIsaacs:
### Describe the changes

Change in `base.js` `set_captcha` of default theme to access data.params.src for displaying the captcha instead of the wrong data.src
Change in `base.js` `set_captcha` of pyplex theme to not convert to json
`parse.bytesize` expectes None if the parameter to convert is bytes instead of "" -> code before leads to traceback
In \_\_init\_\_.py the CaptchaTask does no longer dump the data to string (so overall 2 conversions to json and back are removed)
In captcha.py the recognize (captcha) function doesn't throw a not implemented error anymore, it just returns None
In plugin.py deoding the html file was removed. The simple cast to str made the captcha image an impossible to encode string afterwards, sideeffect of this is that dump_html now has to handle str and bytes for saving to disk (via if isinstance)
In captcha.py the img variable to display later on is now of type bytes, thus has to be decoded before saved into the params dict (line 118)

Different topic:
On loading a ul.to link a call to wait() in hoster.py is done, this also lead to a traceback immediately, beacise in line 356 seconds was none and not comparable to sef.config.get(...), I added a seconds=0 statement for the case that wait is called without any parameters. 

Captchas with hoster nitroflare (e.g.: https://nitroflare.com/view/2AA8ABCD4F9050D) didn't show up, especially in default theme


OzzieIsaacs:
Problem is meanwhile solved

PR #3754, opened 2020-09-20
[Main] Fix uploaded a bytes-like object is required, not 'str'

OzzieIsaacs:
### Describe the changes

After having applied the fix to wait routine another issue raises. I encode the returned html, I only tested it with a dead link, so I'm not sure if there are any other problems during download

```
[2020-09-20 07:05:48]  WARNING             pyload  Download failed: d541h1ys | a bytes-like object is required, not 'str'
  File "/home/ozzie/.local/lib/python3.8/site-packages/pyload/core/threads/download_thread.py", line 56, in run
    pyfile.plugin.preprocessing(self)
  File "/home/ozzie/.local/lib/python3.8/site-packages/pyload/plugins/base/hoster.py", line 306, in preprocessing
    return self._process(*args, **kwargs)
  File "/home/ozzie/.local/lib/python3.8/site-packages/pyload/plugins/base/downloader.py", line 90, in _process
    self._setup()
  File "/home/ozzie/.local/lib/python3.8/site-packages/pyload/plugins/base/hoster.py", line 177, in _setup
    self.grab_info()
  File "/home/ozzie/.local/lib/python3.8/site-packages/pyload/plugins/base/hoster.py", line 238, in grab_info
    new_info = self.get_info(self.pyfile.url, self.data)
  File "/home/ozzie/.local/lib/python3.8/site-packages/pyload/plugins/base/simple_downloader.py", line 148, in get_info
    info.update(cls.api_info(url))
  File "/home/ozzie/.local/lib/python3.8/site-packages/pyload/plugins/downloaders/UploadedTo.py", line 68, in api_info
    api = html.split(",", 4)
TypeError: a bytes-like object is required, not 'str'
```

### Additional references

Tested on Linux Mint 20, python3.8


vuolter:
We have to check why `get_url` returns bytes.

OzzieIsaacs:
> We have to check why get_url returns bytes.

During the call to get_url the load routine in core.network.http_request is called, and this function has a default of "decode=false", which causes the behavior. By passing decode=True to the call to get_url this can be changed without causing problems elsewhere.

vuolter:
I think we have to investigate a little about this `decode=True` and when it should be actually set `true` now that we switched to unicode everywhere.

Issue #3758, opened 2020-09-21
File.Al support


igel-kun:
### Description

FileAl (http://file.al) seems to be an `XFSHoster`-type hoster. I have a [plugin in my fork](https://github.com/igel-kun/pyload/blob/stable/module/plugins/hoster/FileAl.py) that used to work until recently, but they switched to RecaptchaV2 and the current interactive captcha implementation seems to have trouble with it (interactive window comes up in the web-interface, but the captcha doesn't load). Maybe you have a better shot at writing the plugin or figuring out what's up...

### Sample resources

sample link: https://file.al/hcfxima0o9mo/VA_-_Hit_Music_2016_Vol_2_MP3~320Kbps.part2.rar.html

disclaimer: I have nothing to do with this link, I just found it googling for file.al samples.

GammaC0de:
Plugin still works, 
you have to install [Ignore X-Frame-Options](https://addons.mozilla.org/en-US/firefox/addon/ignore-x-frame-options-header) plugin to make captcha work.

igel-kun:
awesome, thanks for the quick support. I'll drop a pull request for the plugin then :] *cheers*

PR #3759, opened 2020-09-21
Create FileAl.py

igel-kun:
### Describe the changes

Interfaces with the http://file.al hoster

see #3758

### DIFF
--- a/module/plugins/hoster/FileAl.py
+++ b/module/plugins/hoster/FileAl.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+#
+# ATTENTION: if you cannot see the interactive captcha (on firefox), make sure to activate/install X-Frame-Options Header:
+# https://addons.mozilla.org/en-US/firefox/addon/ignore-x-frame-options-header/
+
+from module.plugins.internal.XFSHoster import XFSHoster
+import re
+
+class FileAl(XFSHoster):
+    __name__    = "FileAl"
+    __type__    = "hoster"
+    __version__ = "0.01"
+    __status__  = "testing"
+
+    __pattern__ = r'https?://(?:www\.)?file\.al/\w{12}'
+
+    __description__ = """File.al hoster plugin"""
+    __license__     = "GPLv3"
+    __authors__     = [("igel", "")]
+
+
+    PLUGIN_DOMAIN = "file.al"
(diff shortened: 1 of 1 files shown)


module/plugins/hoster/FileAl.py
```diff
@@ -0,0 +1,32 @@
...
+    __license__     = "GPLv3"
+    __authors__     = [("igel", "")]
+
+    PLUGIN_DOMAIN = "file.al"
+    LINK_PATTERN     = r'direct link.*?<a [^>]*href="(.+?)".*?>Click here to download', re.MULTILINE | re.DOTALL
+    WAIT_PATTERN     = r'countdown.*?seconds.*?(\d+)'
+    # "extend" the XFSHoster dict
+    RECAPTCHA_PATTERN = r"g-recaptcha.*?sitekey=[\"']([^\"]*)"
+    PREMIUM_ONLY_PATTERN  = r'(?:[Pp]remium Users only|can download files up to.*only)'
+    LOCAL_LOCATION = r'check_ip'
```
GammaC0de, line 28: What is LOCAL_LOCATION?

igel-kun: Ah, you're right, I forgot I included that in `XFSHoster`: sometimes, fileal would answer with a `location=` in the header while still advancing in the `op`-chain (I don't remember when exactly or why it did that). With the standard `XFSHoster`, this leads to pyload trying to download the HTML-file given in the `location=` header field, which is not what we want. Since, for FileAl, the `location=...` header field would contain the substring `check_ip` somewhere, I decided to check for this substring and, if found, to not use it as direct link but to follow it in order to advance in the `op`-chain. I figured this may happen for other XFSHoster as well, so I included the code for that in `XFSHoster.py`.

This having said, I'm not sure whether it is still ncessary or not. Do you want me to re-submit without `LOCAL_LOCATION` (seeing that changes to `XFSHoster.py` are probably not allowed at this point)?

GammaC0de:
Please remove LOCAL_LOCATION since `FileAl` worked fine with the current `XFSHoster` so I don't see the point of it.

igel-kun:
> Please remove LOCAL_LOCATION since `FileAl` worked fine with the current `XFSHoster` so I don't see the point of it.

you got it :]

GammaC0de:
Thanks 👍

PR #3763, opened 2020-09-30
[DownsterNet] Plugin improvements and fixes

ill1boy:
### Describe the changes

The api from the hoster downster.net has slightly changed. Also I added a so called flow ID for better error identification and support. The plugin is now more stable and resilient.

The plugin wasn't working properly or reliable regarding authentication at service api.

### Additional references

None

### DIFF
--- a/module/plugins/accounts/DownsterNet.py
+++ b/module/plugins/accounts/DownsterNet.py
@@ -25,7 +25,7 @@ class DownsterNet(MultiAccount):
     __license__ = "GPLv3"
     __authors__ = [(None, None)]
 
-    API_URL = "http://localhost:8081"
+    API_URL = "https://downster.net/api"
 
     def rnd(self):
         return ''.join([random.choice(string.ascii_lowercase + string.digits) for n in xrange(5)])
--- a/module/plugins/accounts/DownsterNet.py
+++ b/module/plugins/accounts/DownsterNet.py
@@ -1,79 +1,96 @@
 # -*- coding: utf-8 -*-
 
-import time
-
-import pycurl
-from module.network.HTTPRequest import BadHeader
+import pycurl, random, string
+from time import time, mktime
+from datetime import datetime
 
 from ..internal.misc import json
 from ..internal.MultiAccount import MultiAccount
+from ...network.HTTPRequest import BadHeader
 
 
 class DownsterNet(MultiAccount):
     __name__ = "DownsterNet"
     __type__ = "account"
-    __version__ = "0.01"
+    __version__ = "0.2"
     __status__ = "testing"
 
-    __config__ = [("mh_mode", "all;listed;unlisted", "Filter hosters to use", "all"),
-                  ("mh_list", "str", "Hoster list (comma separated)", ""),
-                  ("mh_interval", "int", "Reload interval in hours", 12)]
+    __config__ = [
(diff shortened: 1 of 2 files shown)


ill1boy:
How critical are those errors? Is it a requirement to fix them before the pr can be merged?
PS: 80 chars is a bit old nowadays^^

GammaC0de:
Can you explain why a new request for every API call?

ill1boy:
Because I want to have a new flow ID per request and I don't want to have a redundant api_request function. So when using the req variable of the account class in the hoster class, the account class throws an exception when changing the headers while a download is running.
That's why I create a new request for every api call.

GammaC0de:
> Do you have a suggestion how to do it better? I'm not an expert with your software :)

I suggest you tell me what has to be changed from the current plugin and I will try to do it.
of-course I would need a premium account for testing.

ill1boy:
Couldn't you tell me how to do it better and I do it?^^

As you may see from the PR I want all requests to use the api_request function to have the proper headers including the flow_id which is generated from the flow_id function. Also there was a change to the premiumUntil field which is now a utc datetime `datetime.strptime(payload['data']['premiumUntil'], '%Y-%m-%dT%H:%M:%S.%f+00:00')`

GammaC0de:
> Couldn't you tell me how to do it better and I do it?^^

You made so many changes to the code I don't know where to start..
For example:
All the work is moved to the `signin()` function,  removing the` grab_info()`  function from the plugin.
creating a new request for every API call, overriding `close()` function

> I don't want to have a redundant api_request function

You mean you  don't want to have a duplicate api_request function?

> the account class throws an exception when changing the headers while a download is running.

Probably because you are using the accounts request for the hoster plugin and the request is currently busy - I need an account to be sure.

P.S.
Please don't fell offended from all of me remarks, I just try to help to make a quality plugin.

P.P.S.
Do you want to try to fix those issues in the plugin?

ill1boy:
Hehe no worries. Just want to learn and improve :) If you want to do it it's fine for me too as it's your software :)

ID: pyload@test.com
PW: pyload

The account will get traffic after 00:00

GammaC0de:
OK,
I implemented the changes you requested, please test and tell if you have any remark.

> premiumUntil field which is now a utc datetime

What is the reason have you changed it?
Now you have to deal with timezone conversion which python sucks with if you don't install the tzinfo package

ill1boy:
It had several reasons. We are using UTC only so +00:00 all the time so it should not be a problem I guess. And if it's an issue I doubt there will be a user trying to use the account within it's last hours :)

Regarding your changes, I like the class based approach and it looks ok to me. Only tiny remark is the function must not be called api_request when the variable is already named api. `self.api.request('user/info')` would be enough I guess. 
Anyways, Thank you for your support :)

GammaC0de:
> Only tiny remark is the function must not be called api_request when the variable is already named api. self.api.request('user/info') would be enough I guess.

Done: https://github.com/pyload/pyload/commit/acd446ed949b6293b2c47f32b38bf3641bc28b5a

ill1boy:
Hey, a user is getting an error when downloading.
`failed 'DownsterApi' object has no attribute 'request'`

GammaC0de:
Make sure he is using account plugin version 0.04 and hoster plugin version 0.03

Restarting pyload might fix this issue.

If still not working, please open an issue and provide debug log.

PR #3764, opened 2020-10-02
Added Letsupload.org downloader plugin (not working)

indestructable-mushu:
So I added a plugin for [letsupload.org](https://letsupload.org/) and it works in the main branch if the encoding guessing code is removed from src/pyload/core/network/http/http_request.py

        encoding = "utf-8"  #: default encoding

        #for line in header:
        #    line = line.lower().replace(b" ", b"")
        #    if not line.startswith(b"content-type:") or (
        #        b"text" not in line and b"application" not in line
        #    ):
        #        continue

        #    none, delemiter, charset = line.rpartition(b"charset=")
        #    if delemiter:
        #        charset = charset.split(b";")
        #        if charset:
        #            encoding = charset[0]

The reason you have to remove it is because it incorrectly parses the html page with _Content-Type: application/octet-stream_ as binary instead of UTF-8 and it's sorta fixed by 36c63c6 + 0e7f47d5 but _application/x-7z-compressed content type_ still doesn't work.

Anyway, I'm having trouble getting the develop branch to work even after disabling the encoding code above. When I try to download any file from letsupload, I get the error message below but it weirdly doesn't show up when debugging. Commenting out _self.c.execute_ just makes the download jump between the filename and url and never starting.

    pyload/core/database/file_database.py", line 244, in update_link
    sqlite3.InterfaceError: Error binding parameter 4 - probably unsupported type.

    @style.async_
    def update_link(self, f):
        self.c.execute(
            "UPDATE links SET url=?,name=?,size=?,status=?,error=?,package=? WHERE id=?",
            (f.url, f.name, f.size, f.status, f.error, str(f.packageid), str(f.id)),
        )

When debugging, it randomly crashes at different parts (ex below) and it even sometimes downloads the file successfully so its likely a race condition problem but I don't know how to debug it. Any idea what's wrong?

        File "pyload/plugins/base/hoster.py", line 248, in check_status
            status = self.pyfile.status
        AttributeError: 'NoneType' object has no attribute 'status'

Sample letsupload files:
https://letsupload.org/cm3p (16kb compiled C file that prints "Hello")
https://letsupload.org/8a65 (10MB zip file)

### DIFF
--- a/src/pyload/plugins/base/simple_decrypter.py
+++ b/src/pyload/plugins/base/simple_decrypter.py
@@ -133,7 +133,7 @@ class SimpleDecrypter(BaseDecrypter):
 
         if "N" in info["pattern"]:
             name = replace_patterns(info["pattern"]["N"], cls.NAME_REPLACEMENTS)
-            info["name"] = parse_name(name)
+            info["name"] = name
 
         return info
 
--- a/src/pyload/plugins/base/simple_downloader.py
+++ b/src/pyload/plugins/base/simple_downloader.py
@@ -194,7 +194,7 @@ class SimpleDownloader(BaseDownloader):
 
         if "N" in info["pattern"]:
             name = replace_patterns(info["pattern"]["N"], cls.NAME_REPLACEMENTS)
-            info["name"] = parse_name(name)
+            info["name"] = name
 
         if "S" in info["pattern"]:
             size = replace_patterns(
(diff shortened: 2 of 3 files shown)


OzzieIsaacs:
I think you are just to impatient, increasing the waiting time to 5 seconds made it for me:
https://github.com/OzzieIsaacs/pyload/tree/letsupload
The other changes (starting from current develop branch): 
- added import parse_name in simple_download
- converted the regexes for scan_download to binary to make it work
- added string conversion of error parameter to file_database

OzzieIsaacs:
I also would like to recommend the use of self.wait instead of time.sleep for waiting (so the waiting time is visible in the User interface)

indestructable-mushu:
Sorry for the late reply, I didn't get a notification. I just tried your branch and can confirm it works. I'll close this pull request.

Issue #3702, opened 2020-08-16
[Turbobit.Net] Links in form "turb.to" are not recognized and therefore fail to download


OzzieIsaacs:
### Describe the problem

If adding turbobit download links like the following:

https://turb.to/g6rbl37jywbt/SEAL.Team.S03E11E12.Todesschwadron.-.Hinterhalt.German.DD51.Dubbed.DL.720p.AmazonHD.h264-GDR%20.part1.rar.html

They are not recognized as turbobit.net but as "XFileSharing" link

![grafik](https://user-images.githubusercontent.com/6438895/90337490-20dac580-dfe3-11ea-9fdf-2074547c98df.png)

### Debug log

```
16.08.2020 17:11:58 INFO      Download starts: SEAL.Team.S03E11E12.Todesschwadron.-.Hinterhalt.German.DD51.Dubbed.DL.720p.AmazonHD.h264-GDR .part1.rar.html
16.08.2020 17:11:58 DEBUG     ADDON ExternalScripts: No script found under folder `download_preparing`
16.08.2020 17:11:58 DEBUG     ADDON UserAgentSwitcher: Setting connection timeout to 60 seconds
16.08.2020 17:11:58 DEBUG     ADDON UserAgentSwitcher: Setting maximum redirections to 10
16.08.2020 17:11:58 DEBUG     ADDON UserAgentSwitcher: Use custom user-agent string `Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:71.0) Gecko/20100101 Firefox/71.0`
16.08.2020 17:11:58 DEBUG     HOSTER XFileSharing[9]: TurbTo | Plugin version: 0.66
16.08.2020 17:11:58 DEBUG     HOSTER XFileSharing[9]: TurbTo | Plugin status: testing
16.08.2020 17:11:58 WARNING   HOSTER XFileSharing[9]: TurbTo | Plugin may be unstable
16.08.2020 17:11:58 INFO      HOSTER XFileSharing[9]: TurbTo | Grabbing link info...
[...]

PR #3770, opened 2020-10-11
[Main] Recognize ddl.to and turb.to Issue 3702

OzzieIsaacs:
### Describe the changes

I applied my suggested change from issue #3702 to make XFelesharing last in the list to compare the regexes
I furthermore updated ddl and turbobit plugin

Issue #3702

### DIFF
--- a/src/pyload/core/managers/plugin_manager.py
+++ b/src/pyload/core/managers/plugin_manager.py
@@ -6,8 +6,8 @@ import re
 import sys
 from ast import literal_eval
 from itertools import chain
+from collections import OrderedDict
 
-# import semver
 
 from pyload import APPID, PKGDIR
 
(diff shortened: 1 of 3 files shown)


GammaC0de:
Please remove the changes to plugin_manager.py

In my opinion, the filename should not affect the  plugin order.
Also, I intend to make Xfilesharing disabled by default

OzzieIsaacs:
Hello @GammaC0de, I'm not ready to give up yet. I agree with you the name is not the best solution to get the plugins ordered. So another proposal:
I've added a member \_\_order\_\_, it can have any number (integer would normally sufficient, I now used float to have the same regex as for the version parser). 
This \_\_order\_\_ variable is added to the dict and used for sorting. The only plugin which needs the order member really set is the XFIleSharing plugin to move it to the end, everything else can stay at default (e.g. 0).

What about this (I already changed the PR accordingly)?

vuolter:
As I remember plugins with pattern `^unmatchable$` should already run at the end, if no one else is matched. If it does not happen, there is a bug or some code was accidentally removed during the migration to py3.

OzzieIsaacs:
@voulter: I checked it again, it's not working (not for the development branch and also not for stable, as reported in #3702). 
I didn't 100% get what is going on. It's matching with the name of XFilesharing, but the pattern looks like the one from "MultiAccount" (at least in pyload_ng)

PR #3777, opened 2020-10-17
[New Hoster] Userupload Net

OzzieIsaacs:
### Describe the changes

Added new Hoster Userupload.Net

Sample link:
https://userupload.net/t88dwntmmy1i


GammaC0de:
Can't it be done without a custom `handle_free()`?

Also, `search_pattern()` is unnecessary here, you simply can use `re.search()`

OzzieIsaacs:
I changed search pattern ti re.search
I created a handle_free because I could not get the filename from the website, there is only a trunkated version available (the sample has a very long filename)
I also had problems to make the regex match with the "re.S" flag applied

PR #3791, opened 2020-10-24
[Main] Updated all bool int config get routines

OzzieIsaacs:
### Describe the changes

Your decision to reject my patch for #3736 still leads to the situation:

Currently all config settings of type bool and int are converted to strings, whereas all (**except one** -> autologin) requests to the settings are expecting bool values. 

Your decision to check the return value against "True" in the calling function now leads to the situation that all calls to a config routine with bool or int type have to convert the string to a bool/int value by itself (I'm not happy with this decision, but it's you project)
This patch addresses this issue by changing 17 files with 30+ lines changed instead of the 2 lines added by my patch.

All bool/int settings are expected in the code to be bool or int. You can take each variable and check if it works correct, it will not.

### DIFF
--- a/src/pyload/core/log_factory.py
+++ b/src/pyload/core/log_factory.py
@@ -53,9 +53,9 @@ class LogFactory:
         return logger
 
     def _init_logger(self, logger):
-        console = self.pyload.config.get("log", "console")
-        syslog = self.pyload.config.get("log", "syslog")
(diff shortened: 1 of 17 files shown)


GammaC0de:
> I'm not happy with this decision, but it's you project

I have decided to return the real value, not string.  https://github.com/pyload/pyload/commit/d8881bf982f56d72dd1a53bceabd591e72e13cc2

PR #3745, opened 2020-09-19
[Main] Fix traceback on change user

OzzieIsaacs:
### Describe the changes

Typecast of Perms enum to str to use startswith function

### Is this related to a problem?

If you go to admin page and try to change the users permissions you get an traceback:
```
[2020-09-18 17:49:59]  INFO                pyload  ADDON ClickNLoad: Proxy listening on 0.0.0.0:9666
[2020-09-18 17:51:02]  DEBUG         pyload.webui  'Perms' object has no attribute 'startswith'
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/ozzie/.local/lib/python3.8/site-packages/pyload/webui/app/helpers.py", line 181, in wrapper
    response = func(*args, **kwargs)
  File "/home/ozzie/.local/lib/python3.8/site-packages/pyload/webui/app/blueprints/app_blueprint.py", line 463, in admin
    data["permission"] = set_permission(data["perms"])
  File "/home/ozzie/.local/lib/python3.8/site-packages/pyload/webui/app/helpers.py", line 117, in set_permission
    if name.startswith("_"):
AttributeError: 'Perms' object has no attribute 'startswith
```
This patch fixes this.

### Additional references

There is one more problem, I don't know how you like to handle it. After appling the fix you can change the users permisions. But if you are logged in as the only user pyload and try to remove it's admin rights you get the following traceback (untick the checkbox below admin):
```
[2020-09-19 08:14:50]  DEBUG         pyload.webui  'role'
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/matthias/.local/lib/python3.8/site-packages/pyload/webui/app/helpers.py", line 181, in wrapper
    response = func(*args, **kwargs)
  File "/home/matthias/.local/lib/python3.8/site-packages/pyload/webui/app/blueprints/app_blueprint.py", line 465, in admin
    api.set_user_permission(name, data["permission"], data["role"])
[...]

Issue #3734, opened 2020-09-13
[Main] Improved user management


OzzieIsaacs:
### Feature description

With pyload 4.20 it was possible to add/delete user via the command line. This option is now missing in pyload NG. This feature is essential and could be realized with the Web UI. Whats currently missing in my opinion:

- [ ] Add users

- [ ] Delete user

- [ ] Change passwords of other users as admin (currently you have to know the password of the user you want to change the     password, which is not very clever)

- [ ] Prevent last admin user from removing it's admin rights

- [ ] Optional change username (could be worked around by creating a new user and deleting the old user)

### Problem description
It's not possible to manage additional users

Issue #3732, opened 2020-09-13
[Main] Change userpassword displays wrong hints


OzzieIsaacs:
### Description
If you are using the pyplex theme with the pyload NG and try to change the password you get displayed a wrong message what to do to change the password. The Default theme displays the right message

Current:
To add user or change passwords use: python pyLoadCore.py -u
Important: Admin user have always all permissions!

Should be:
To restore default admin login run: pyload -r
Important: Admin user have always all permissions!"

### Debug log
N/A

### Screenshots

![grafik](https://user-images.githubusercontent.com/6438895/93021141-7e156700-f5e1-11ea-9e99-56e90a223399.png)

PR #3795, opened 2020-10-24
[Main] Enhancement user management

OzzieIsaacs:
### Describe the changes

This PR addresses some problems around the user management.

1) Upon start the database was search in the folder .pyload/pyload.db, wehreas the location is in .pyload/data/pyload.db -> \_\_init\_\_.py

2) Wrong hints for adding users have been removed (pyplex theme) (#3732)

3) Added a button and a dialog box for adding users (also extended the backend towards this) (#3734)

4) Added a button to delete certain users, this button is removed for the current user, to prevent the case the last admin deletes himself. The button is furthermore only visible if the current user has admin rights

5) The applied permissions are shown correct for users (before always nothing was shown)

6) Listing of permissions was changed again, as the first patch only prevented the traceback, but didn't solve the root problem. (#3745)

### Additional references

grafik


OzzieIsaacs:
@GammaC0de Just for your information, your latest patch was no.1 of the fixed problems in this PR

PR #3799, opened 2020-10-24
[Main] Fix decode header

OzzieIsaacs:
### Describe the changes

The header is decoded from a binary header variable by casting the variable to str. This creates a scrappy header variable with content like:
`b'HTTP/2 302 \r\ndate: Sat...PRG\r\n\r\n'`

where the header variables can't be extracted from. The PR uses a .decode('utf-8')  Instead.

The behavior can be observed by using this download link:

https://ddownload.com/w14i1dpjtnay

(The captcha is requested, the link is also retrieved, but not recognized, therefore pyload tries to reload the link, the website knows that we are having the link and let's us 2 hours wait..)

Download with ddownload.org not working.

### DIFF
--- a/src/pyload/plugins/helpers.py
+++ b/src/pyload/plugins/helpers.py
@@ -434,7 +434,7 @@ def set_cookies(cj, cookies):
 
 
 def parse_html_header(header):
-    header = str(header)
+    header = header.decode('utf-8')
 
     hdict = {}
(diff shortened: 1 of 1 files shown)


GammaC0de:
Better use `to_str()`

GammaC0de:
Thanks

PR #3821, opened 2020-10-31
[Main] Fix storagedir not persistent after restart

OzzieIsaacs:
### Describe the changes

If you change the download directory on the web UI and afterwards restart pyload-ng the setting is reset to default. This is caused by the arg parser giving the default storage dir as parameter to the core init function. In the first step the storage dir is saved to the config file. The patch sets the default argument to none and checking in init if some argument is present, if not, the storagedir is loaded from the config file  

Storage dir is not persitant across restarts


vuolter:
Nice work 👍

PR #3822, opened 2020-10-31
[Main] Make path_prefix for webserver working

OzzieIsaacs:
### Describe the changes

The current implementation for the webserver path_prefix isn't working at all. To make it work with cheroot the pathprefix should be "/path_prefix" and not "path_prefix/" as now. Doing so would create problems, as the jinja templating engine isn't aware of where it is currently sitting ("/" or "/path_prefix"), so all links from the html and js files are pointing to the wrong location. All of the links are hardcoded, which is alos not a good idea.
This PR adds a url_prefix parameter for all blueprints upon registering. The current url_prefix in the blueprints are moved to the route decorator (e.g. "/json/status" instead of "/status"). The theme init is seperated from the other extension to also add a ulr_prefix in the form of "/url_prefix/_theme")
All routes in the html fies are generated using the "url_for" function. The json and api routes in the js files are now also generated using url_for. 

### Additional references

I only did it for the pyplex theme, the other themes have to be done if you are agree to this change

### DIFF
--- a/src/pyload/webui/app/extensions.py
+++ b/src/pyload/webui/app/extensions.py
@@ -8,12 +8,14 @@ from pyload import APPID
 
 
 class Themes(_Themes):
-    def init_app(self, app):
-        return super().init_themes(app, app_identifier=APPID)
+    def init_app(self, app, path_prefix=""):
+        return super().init_themes(app, app_identifier=APPID, theme_url_prefix=path_prefix + "/_themes")
 
 
 babel = Babel()
 cache = Cache()
 themes = Themes()
 
-EXTENSIONS = [babel, cache, themes]
+EXTENSIONS = [babel, cache]
+
+THEMES = [themes]
--- a/src/pyload/webui/app/extensions.py
+++ b/src/pyload/webui/app/extensions.py
@@ -9,7 +9,9 @@ from pyload import APPID
 
 class Themes(_Themes):
     def init_app(self, app, path_prefix=""):
-        return super().init_themes(app, app_identifier=APPID, theme_url_prefix=path_prefix + "/_themes")
+        return super().init_themes(app,
+                                   app_identifier=APPID,
+                                   theme_url_prefix=path_prefix + "/_themes")
 
 
 babel = Babel()
--- a/src/pyload/webui/app/templates/js/dashboard.js
+++ b/src/pyload/webui/app/templates/js/dashboard.js
@@ -10,7 +10,7 @@ document.addEvent("domready", function(){
 var EntryManager = new Class({
     initialize: function(){
         this.json = new Request.JSON({
-            url: "{{url_for("json.links")}}",
+            url: "{{url_for('json.links')}}",
             secure: false,
             async: true,
             onSuccess: this.update.bind(this),
@@ -163,7 +163,7 @@ var LinkEntry = new Class({
 
         this.elements.remove.addEvent('click', function(){
             new Request({
-                method: 'get', url: '/json/abort_link/' + this.id
+                method: 'get', url: window.location.pathname + "/../json/abort_link/" + this.id
             }).send();
         }.bind(this));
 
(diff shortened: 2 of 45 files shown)


src/pyload/webui/app/extensions.py
```diff
@@ -8,12 +8,14 @@
...
 
 
 babel = Babel()
 cache = Cache()
 themes = Themes()
 
-EXTENSIONS = [babel, cache, themes]
+EXTENSIONS = [babel, cache]
+
+THEMES = [themes]
```
GammaC0de, line 21: Please add newline

GammaC0de:
Very nice :+1:

Please update the other themes

OzzieIsaacs:
I updated the other themes. 
One new question came up? What to do with the cnl load server: If you have a path prefix shall it also listen on it (e.g. 127.0.0.1:9666/test/flash/... instead of  127.0.0.1:9666/flash/..), or shall it be unchanged (?)

This change is now not included in #3795, #3737 and #3744. Would be great if you could be give feedback regarding this ones, I'll start merging and checking the PRs afterwards

src/pyload/webui/app/templates/js/dashboard.js
```diff
@@ -163,7 +163,7 @@ var LinkEntry = new Class({
 
         this.elements.remove.addEvent('click', function(){
             new Request({
-                method: 'get', url: '/json/abort_link/' + this.id
+                method: 'get', url: window.location.pathname + "/../json/abort_link/" + this.id
```
GammaC0de, line 166: Shouldn't this be:
`method: 'get', url: "{{url_for('json.abort_link')}}"  + this.id`?

src/pyload/webui/app/templates/js/packages.js
```diff
@@ -107,7 +107,7 @@ var PackageUI = new Class({
             indicateLoad();
             new Request.JSON({
                 method: 'get',
-                url: '/json/package_order/' + order[0],
+                url: window.location.pathname + "/../json/package_order/" + order[0],
```
GammaC0de, line 110: Same thing:
`url: "{{url_for('json.package_order')}}" + order[0],`
no?

src/pyload/webui/app/templates/js/packages.js
```diff
@@ -281,7 +281,7 @@ var Package = new Class({
         indicateLoad();
         new Request({
             method: 'get',
-            url: '/api/delete_packages/[' + this.id + ']',
+            url: window.location.pathname + "/../json/delete_packages/[" + this.id + ']',
```
GammaC0de, line 284: And here

GammaC0de:
All those `window.location.pathname` have to be fixed...

> This change is now not included in #3795, #3737 and #3744. Would be great if you could be give feedback regarding this ones, I'll start merging and checking the PRs afterwards

I had no time to look at it, been busy with life..
I'll check them after we finish with this PR.

Regarding ClickNLoad,
This has to be checked,
As far as I can tell, ClickNLoad should not be affected by path_prefix

OzzieIsaacs:
@GammaC0de 
This is a case I should have mentioned, indeed. All `/json/..` routes with parameters can't be build using the url_for function, because jinja is detecting that a parameter is missing and throws a traceback "route not found, did you mean ..".

There are 4 possibilities to avoid this:

1) Use `window.location.pathname` or similar javascript functions to detect the location of the js file and change the path of the call accordingly (my way)

2) Refactor json_blueprint.py similar to api_blueprint, to have only a generic route input for all routes with parameter and change the calls in js script to something like `"{{url_for('json.routine', func='package_order')}}/[" + id + "]"`, you then loose the parameter type check for all routines, not only the ones with parameter

3) change all `url_for(..)` calls in js to `window.location.pathname` (solution 1), url_for routines in the html code will not be affected by this 

4) Adding a route with a default parameter on every route in json_blueprint.py like:
```
@bp.route("/json/package", defaults={'id': -1}, endpoint="package")
@bp.route("/json/package/<int:id>", endpoint="package")
@login_required("LIST")
```
and handle the calls with invalid ids (`if id == -1: flask.abort(500)`)

What shall we(I) do?

GammaC0de:
Thanks

GammaC0de:
https://github.com/pyload/pyload/commit/a5052a5ed3ad913892f30d4feb487c6e62deae0e
Now all routes uses `url_for()`

PR #3803, opened 2020-10-25
[Main] Migrate Crypto to Cryptography for clicknload

OzzieIsaacs:
Fixed missing parameter thread

### Describe the changes

During debugging the addon_thread is complaing about unexpected keyword argument 'thread'
1) By adding the parameter thread to forward, _server and proxy function in clicknload this can be avoided

2) The clicknload decrypting routine weren't working with Fernet, I could not make it run, (Is fernet and AES the same?). I migrated it to cryptography following your example from dlc.py. I send @GammaC0de an example link via e-mail

### Is this related to a problem?

```
[2020-10-25 10:40:10]  DEBUG         pyload.webui  Fernet key must be 32 url-safe base64-encoded bytes.
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/ozzie/.local/lib/python3.8/site-packages/pyload/webui/app/blueprints/cnl_blueprint.py", line 30, in wrapper
    return func(*args, **kwargs)
  File "/home/ozzie/.local/lib/python3.8/site-packages/pyload/webui/app/blueprints/cnl_blueprint.py", line 108, in addcrypted2
    obj = Fernet(key)
  File "/home/ozzie/.local/lib/python3.8/site-packages/cryptography/fernet.py", line 36, in __init__
    raise ValueError(
ValueError: Fernet key must be 32 url-safe base64-encoded bytes.

```

PR #3831, opened 2020-11-06
[Main] Fix jdcheck.js with cript.to

OzzieIsaacs:
### Describe the changes

Add the expected header to the jdcheck.js response (copied from current jdownloader responses). Furthermore added catch exceptions on forwarded cnl packets.

To make this PR working PR #3803 has to be merged to (otherwise the data can't be decrypted and pyload errors out)

Using cript.to to add links via click n' load leads to an error (on cript.to side). 
This can be checked with this link: `https://cript.to/folder/PD0ps4u4`
Cript seems to check not only the content of the response but also the header (and stopping transfer if the header isn't as expected):
grafik

grafik

Cript.to also seems not to complete the transfer, but just interrupts the connection after sending the payload -> traceback in debugger (Datatransfer interrupted and socket not connected are the error messages)

```
python-BaseException
  File "/home/ozzie/.local/lib/python3.8/site-packages/pyload/plugins/helpers.py", line 541, in forward
    destination.sendall(bufdata)
BrokenPipeError: [Errno 32] Datenübergabe unterbrochen (broken pipe)
During handling of the above exception, another exception occurred:
  File "/home/ozzie/.local/lib/python3.8/site-packages/pyload/core/threads/addon_thread.py", line 50, in run
    self.f(*self.args, **self.kwargs)
  File "/home/ozzie/.local/lib/python3.8/site-packages/pyload/plugins/addons/ClickNLoad.py", line 98, in forward
    forward(source, destination)
  File "/home/ozzie/.local/lib/python3.8/site-packages/pyload/plugins/helpers.py", line 547, in forward
    destination.shutdown(socket.SHUT_WR)
OSError: [Errno 107] Der Socket ist nicht verbunden
```

### DIFF
--- a/src/pyload/plugins/addons/ClickNLoad.py
+++ b/src/pyload/plugins/addons/ClickNLoad.py
@@ -91,7 +91,7 @@ class ClickNLoad(BaseAddon):
 
     @lock
     @threaded
-    def forward(self, source, destination, queue=False):
+    def forward(self, source, destination, queue=False, thread=None):
         if queue:
             old_ids = set(pack.pid for pack in self.pyload.api.get_collector())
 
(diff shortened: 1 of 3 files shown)


GammaC0de:
It seems that cript.to's ClickNLoad implamantation is broken:
it passes the` jcheck..js` check successfully  but then fails to send the `addcrypted2` request, even with `jdownloader`.

image

PR #3837, opened 2020-11-08
[Main] Add missing url_for on logout -> login redirect

OzzieIsaacs:
### Describe the changes

On the logout template there is one redirect route, which wasn't considered with the url_for route generating change for making the path_prefix work. This PR addresses this issue. Furthermore I had a problem on login pointing next to the login path, so I excluded this usecase from the next route generation

Problems on logout -> login with path_prefix

### DIFF
--- a/src/pyload/webui/app/helpers.py
+++ b/src/pyload/webui/app/helpers.py
@@ -30,7 +30,7 @@ def get_redirect_url(fallback=None):
     for location in flask.request.values.get("next"), flask.request.referrer:
         if not location:
             continue
-        if location == flask.request.url:  # don't redirect to same location
+        if location == flask.request.url or flask.request.url.endswith('login'):  # don't redirect to same location
             continue
         if is_safe_url(location):
             return location
--- a/src/pyload/webui/app/templates/logout.html
+++ b/src/pyload/webui/app/templates/logout.html
@@ -1,7 +1,7 @@
 {% extends 'base.html' %}
 
 {% block head %}
-<meta http-equiv="refresh" content="3; url="{{url_for('app.dashboard')}}">
+<meta http-equiv="refresh" content='3; url="{{url_for('app.dashboard')}}"'>
 {% endblock %}
 
 {% block content %}
--- a/src/pyload/webui/app/themes/modern/templates/logout.html
+++ b/src/pyload/webui/app/themes/modern/templates/logout.html
@@ -1,7 +1,7 @@
 {% extends theme('base.html') %}
 
 {% block head %}
-<meta http-equiv="refresh" content="5; url='/'">
+<meta http-equiv="refresh" content='5; url="{{url_for('app.dashboard')}}"'>
 <link rel="stylesheet" type="text/css" href="{{theme_static('css/logout.css')}}" />
 {% endblock %}
 
--- a/src/pyload/webui/app/themes/pyplex/templates/logout.html
+++ b/src/pyload/webui/app/themes/pyplex/templates/logout.html
@@ -1,7 +1,7 @@
 {% extends theme('base.html') %}
 
 {% block head %}
-<meta http-equiv="refresh" content="5; url='/'">
+<meta http-equiv="refresh" content="5; url="{{url_for('app.dashboard')}}">
 <link rel="stylesheet" type="text/css" href="{{theme_static('css/logout.css')}}" />
 {% endblock %}
 


GammaC0de:
> Furthermore I had a problem on login pointing next to the login path, so I excluded this usecase from the next route generation

Can you explain please?

OzzieIsaacs:
@GammaC0de : I don't got it to run with the path prefix (But I thought it would work, so I didn't pay attention to it up to now).
My reverse proxy setup is the following (apache 2.4):

```
Redirect /pyload /pyload/
<Location /pyload>
      # for login preserve host has to be set to true in virtual host 
       order allow,deny
       allow from all
       ProxyPass http://localhost:8000
       Header edit Location ^[^/]*(//[^/]*)(.*)$ $1/pyload$2
       Header edit Set-Cookie "^(.*; Path=)(.*)" $1/pyload$2
       RequestHeader unset Accept-Encoding
       SetOutputFilter html;css;js;jsapp
</Location>

ExtFilterDefine html mode=output intype=text/html cmd="/bin/sed s#src=\"/#src=\"/pyload/#g;s#href=\"/#href=\"/pyload/#g;s#/json/#/pyload/json/#g;s#=\"/pathchooser#=\"/pyload/pathchooser#g;s#url=\"/\"#url=\"/pyload\"#>
ExtFilterDefine css mode=output intype=text/css cmd="/bin/sed s#/media/#/pyload/media/#g"
ExtFilterDefine js mode=output intype=text/javascript cmd="/bin/sed s#/json/#/pyload/json/#g;s#/media/#/pyload/media/#g;s#/api/#/pyload/api/#g;s#/home#/pyload/home#g;s#/filechooser#/pyload/filechooser#g;s#/pathchoose>
ExtFilterDefine jsapp mode=output intype=application/javascript cmd="/bin/sed s#/json/#/pyload/json/#g;s#/media/#/pyload/media/#g;s#/api/#/pyload/api/#g;s#/home#/pyload/home#g;s#/filechooser#/pyload/filechooser#g;s#/>
```
To get the usecase where the next-problem occurs do the following: 
Set `Skip login if single user:` to off
`path prefix` is not needed (has to be empty) with this setup, has the benefit you can access pyload directly from network on the original port and in parallel via the reverse proxy with the path prefix (pyload in my example)
Access pyload via reverse proxy
Restart pyload, keep the browser open
Click on any icon (e.g. dashboard), you get redirected to login page
Enter your credentials -> get logged in, but you see an error 404 page not valid, path is /pyload/pyload/login instead of /pyload/dashboard

This might be not a valid usecase. If I enter path prefix and and try to acces pyload under http://127.0.0.1:8000/pyload
You are stuck in an endless redirect loop from login to dashboard to login and so on. I tried to fix it, but I didn't get it, in the login_required function you never get the flask.session with the information that you are logged in, instead you get a new session... 
(Checked it with python3.8 on Linux Mint with Firefox 82.02)

GammaC0de:
Just to make sure, does it work OK now after https://github.com/pyload/pyload/commit/fb95c3e149b165b06ad0a28354d22678aabc7c71?

OzzieIsaacs:
I didn't merge it before, which I now did. So my solution is still working, you are now redirected to the page you came from after login, which is much better than before.

I found one last problem with the path_prefix config:
Having `develop` set to false, `path_prefix` set to /pyload and `Skip login if single user` to false you can login. Pressing on logout you get to the "you are logged out" page:
grafik
This page is constantly reloading (I guess "next" is set to logout, without checking it..)

OzzieIsaacs:
The html logout page is showing errors:
grafik

I think:
`<meta http-equiv="refresh" content="5; url='{{url_for('app.dashboard')}}'">`
would be better

GammaC0de:
logout.html already has these changes: https://github.com/pyload/pyload/blob/develop/src/pyload/webui/app/themes/modern/templates/logout.html#L4

OzzieIsaacs:
Okay, pyplex is missing it: https://github.com/pyload/pyload/blob/e4f9de55d9050e092df822773d6fbecb8a92e332/src/pyload/webui/app/themes/pyplex/templates/logout.html#L4

GammaC0de:
pyplex has it too https://github.com/pyload/pyload/blob/develop/src/pyload/webui/app/themes/pyplex/templates/logout.html#L4

OzzieIsaacs:
I'm talking about the single quote, double quote mismatch:

<meta http-equiv="refresh" content="  5; url="{{url_for('app.dashboard')}}"    > 
<meta http-equiv="refresh" content='  5; url="{{url_for('app.dashboard')}}"    '>

I can create a PR tomorrow for it, but this solves not the endless logout loop, I think..

GammaC0de:
Done https://github.com/pyload/pyload/commit/b3c89c6641dd8a67d872b074b28c1f90e3f193be

GammaC0de:
> but this solves not the endless logout loop, I think..

For me, no issue after https://github.com/pyload/pyload/commit/b3c89c6641dd8a67d872b074b28c1f90e3f193be

OzzieIsaacs:
I had time to check it, it' wired. I can't get back to the state where I can log in. (I think it was because I did several changes reverted it, was logged in and did not close the browser in between).
I can confirm that everything works well with Chromium (I guess also Chrome). 
BUT using firefox you can't get logged in at all, you always get redirected to the login page.
grafik
(right side firefox, left side chromium)
While dbugging the code the flask session is always "empty" -> no user, therefore I get redirected to login page 

I'm running pyload_ng on the same machine I try to access it on Linux Mint 20, without any proxy in between. The path_prefix is set to "/pyload". I type "127.0.0.1/pyload" to get to the login page (right after opening the firefox browser). I deactivated all ad-ons and deleted all cookies from 127.0.0.1, but still the same behavior

GammaC0de:
I am using Firefox without issue (no path prefix though)

PR #3843, opened 2020-11-14
[Main] Ignorecase on match plugins

OzzieIsaacs:
### Describe the changes

Currently the regex to find the right plugin is case sensitive, so for example taking the filecrypt decrypter plugin:

`r"https?://(?:www\.)?filecrypt\.cc/Container/\w+"`

This would match: `http://filecrypt.cc/Container/64E039F859.html`
But not: `http://filecrypt.cc/container/64E039F859.html` (C in container not a capital letter)
or: `http://Filecrypt.cc/Container/64E039F859.html` (Capitalized URL)
or even: `HTTP://Filecrypt.cc/Container/64E039F859.html` (Capitalized HTTP)

I added an "ignorecase" flag to the pattern match regex in pluginmanager, this allows us to match all above examples

### DIFF
--- a/src/pyload/core/managers/plugin_manager.py
+++ b/src/pyload/core/managers/plugin_manager.py
@@ -204,7 +204,7 @@ class PluginManager:
                     plugins[name]["pattern"] = pattern
 
                     try:
-                        plugins[name]["re"] = re.compile(pattern)
+                        plugins[name]["re"] = re.compile(pattern, re.I)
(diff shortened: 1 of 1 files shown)


vuolter:
Thanks :)

GammaC0de:
This has to be solved differently:
only the scheme and host name are case insensitive in a URL, the path and query are case sensitive.

PR #3864, opened 2020-12-22
Fix password changing and checking

kepi:
There was error at least in `change_password` because of missing salt at beginning
of newpw.

IMHO it is better when `_salted_password` returns password including salt and all
methods are adjusted to it.

I also made little cleanup - isolated salt generation and password checking to
own methods.


GammaC0de:
Nice work but please issue the PR against the develop branch,
thanks

kepi:
Oh my... I was so happy I finally got the ng running I didn't realize I'm using outdated code :wink: 

Rebased on develop

PR #3875, opened 2021-01-03
Premium download failed, not mentioning the real cause

benroe89:
The creation of the download folder was not working, because of missing permissions.

But, this was not mentioned in the log.

The try-except clause was passing the Exception.message property to self.fail(), which was empty in that case, as not all exception subclasses, like OSError, have that defined.

### Describe the changes

To solve this the exception is passed calling str() before. Thus it just creates a string out of the exception's arguments.

The (premium) download failed, as stated on the log page of pyload, with the following line:
`HOSTER DdlTo[6207]: Premium download failed | downloading`

As you can see, the cause for the failure is not mentioned, as the current status line is printed there ("downloading")

### DIFF
--- a/module/plugins/internal/Hoster.py
+++ b/module/plugins/internal/Hoster.py
@@ -297,7 +297,7 @@ class Hoster(Base):
                 os.makedirs(dl_dir)
 
             except Exception, e:
-                self.fail(str(e))
+                self.fail(e.args[0])
 
         self.set_permissions(dl_dir)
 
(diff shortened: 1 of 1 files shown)


module/plugins/internal/Hoster.py
```diff
@@ -297,7 +297,7 @@ def download(self, url, get={}, post={}, ref=True, cookies=True,
                 os.makedirs(dl_dir)
 
             except Exception, e:
-                self.fail(e.message)
+                self.fail(str(e))
```
GammaC0de, line 300: Better use `self.fail(e.args[0])`

GammaC0de:

GammaC0de:
Thanks for noticing 👍

Issue #3872, opened 2021-01-02
ExtractArchive - Don't extract the following folder?


uysct:
This setting **Don't extract the following files: *.jpg,*.nfo,*.url,*.DS_Store,index.dat,thumb.db**  does not extract the specified files after a successful download.
Is it possible that certain folders contained in the archive (for example a folder "Sample") are not unzipped?

How can I define that a complete folder with its content isn't touched while extraction? In my specific example I want to exclude the folder sample with all its content. Thanks

rootsradics:
I'm not sure, but I think there's a minor issue with extractors plugin. To fix this one, I had to install 7zip, and add the '*.zip' extension to be handled by 7zip. This way, that works!
The other solution is to add a few lines to the UnZip plugin (extractall part, members have to be passed to the functionà, because AFAIK this one does not handle exclusion (files or directories). 
If I have time, I will try to fix and propose this fix to the pyload team.

GammaC0de:
I can confirm that 7zip plugin should work by assing the folder name to the exclusion list (`Sample` in your example).

uysct:
> The other solution is to add a few lines to the UnZip plugin (extractall part, members have to be passed to the function, because AFAIK this one does not handle exclusion (files or directories).
> If I have time, I will try to fix and propose it to the pyload team.

This is a really great idea! Personally, I like the second solution better, because the included "ExtractArchive" plugin works great for me. Unfortunately, the desired function is missing. The developers will have to decide which of the two suggestions is correct. If you can write code that works, it would be fantastic! Many Thanks

uysct:
> I can confirm that 7zip plugin should work by assing the folder name to the exclusion list (`Sample` in your example).

As a developer, you will surely find the most sensible solution, I hope your time will allow it! Many Thanks

GammaC0de:
> As a developer, you will surely find the most sensible solution, I hope your time will allow it! Many Thanks

It should work fine without any updates if you use 7zip with pyload instead of unrar/unzip.

uysct:
Thank you for your feedback. Unfortunately, I am unable to do this because I simply don't know how to do it. I use pyload in a Docker (writl / pyload) in which this function (unrar / unzip) is already integrated. Is there an easy way to install 7zip?

GammaC0de:
Please enable debug mode and then restart pyload.
You should see a line says something like:
```
[...]

PR #3879, opened 2021-01-07
UnZip Remove files based on exclusion list (#3872)

rootsradics:


src/pyload/plugins/extractors/UnZip.py
```diff
@@ -80,7 +81,11 @@ def extract(self, password=None):
         try:
             with zipfile.ZipFile(self.filename, "r") as z:
                 z.setpassword(password)
-                z.extractall(self.dest)
+                members = (member for member in z.namelist()  
+                           if not any(re.compile(fnmatch.translate(exclusion)).match(member)
+                           for exclusion in self.excludefiles)
```
GammaC0de, line 86: Why not just:
`if not any(fnmatch.fnmatch(member, exclusion)`
?

src/pyload/plugins/extractors/UnZip.py
```diff
@@ -80,7 +81,11 @@ def extract(self, password=None):
...
+                          )
+                z.extractall(self.dest,members=members)
                 self.files = [os.path.join(self.dest, _f)
                               for _f in z.namelist()
                               if _f[-1] != os.path.sep]
```
GammaC0de, line 91: Should be changed to:
`if _f[-1] != os.path.sep and _f in members]`

GammaC0de:
Also, plugin version should be increased.

rootsradics:
Should be OK too !
Sorry, I'm quite new to Python, and I'm missing some API knowledge :)
I will try to pull request with the fix you proposed ( the one below,
and the two others (bump version and add members to the returned files list)

Le 07/01/2021 à 23:08, GammaC0de a écrit :

PR #3999, opened 2021-08-01
A few modifications to fix the Docker image

raycheung:
### Describe the changes

This is to make the Dockerfile build the image. These are minimal changes to pass the build.

The current Dockerfile is flawed and it would not build the Docker image.


vuolter:
I'm sorry for the long delay, looks all okay for me, thank you @raycheung 😃

PR #4019, opened 2021-09-27
Fix externalscripts

wweich:
### Describe the changes

Added `self.pyload.userdir` directory prefix when searching scripts and changed `Popen` argument from generator to list.

I'm have little knowledge of Python and don't no if this is the correct solution, but it works for me.

ExternalScripts did not work for me on Ubuntu 20.04 with Python 3.8

The scripts were searched in ~/scripts instead of ~/.pyload/scripts.
After fixing that, the addon logged "list index out of range" when trying to run the script.

### DIFF
--- a/src/pyload/plugins/addons/ExternalScripts.py
+++ b/src/pyload/plugins/addons/ExternalScripts.py
@@ -87,7 +87,7 @@ class ExternalScripts(BaseAddon):
 
         for folder in self.folders:
             scripts = []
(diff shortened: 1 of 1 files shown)


GammaC0de:
Thanks for your contribution, but you should have used the `develop` branch to submit the PR.
Please submit the PR against the `develop` branch.

GammaC0de:
OK, I have merged this into the develop branch, thanks for you contribution 👍

PR #4041, opened 2021-12-01
Use only HTTPS for MultiUp.org

kepi:
MultiUp.org is now (for couple months already) redirecting all request to https.
It is better to directly use https than depend on redirects, which might bring
more errors.

### Describe the changes

Just changed addresses from http://www.multiup.org to https://www.multiup.org

Yes, in my case this was required couple months ago (I forgot to send PR) to work. Not sure if it is still needed, maybe redirect is handled correctly now, but it is better in any case to use final address than depend on redirect on their side.


GammaC0de:
Please increase __version__

kepi:
Sorry, increased.

PR #4055, opened 2021-12-24
Fix pycurl issue when running in docker

raycheung:
### Describe the changes

Fix pycurl issue when running in docker

The pycurl has confused across 2 different libcurl versions for compilation-time & run-time.


GammaC0de:
Thanks 👍

PR #4068, opened 2022-01-09
Improve docker-compose example

fmirkes:
### Describe the changes

- Fix version string
- Added  Click 'N' Load port
- Use local Dockerfile instead of docker hub image (currently not available)

- the current `version` is an integer, which prevents docker-compose from running
- there are no pyload images available on [docker hub](https://hub.docker.com/r/pyload/pyload/tags)


GammaC0de:
👍

PR #4075, opened 2022-01-18
Automatic addition of passwords via Click 'n' Load plugin.

ajayvarghese2000:
### Describe the changes

Added a method to the click 'n' load plugin that adds the password to a package, if a password was supplied for purposes such as archive extractions etc.

Was trying to add a bunch of archives that are password protected using a click 'n' load button. Although the links and file name was passed through perfectly fine, the password was ignored.

After looking at the `cnl_app.py` file it was clear to see that no method was implemented in dealing with passwords so I added it it.

### Additional references

The click 'n' load data must follow the [JDownloader cnl2 standard](https://jdownloader.org/knowledge/wiki/glossary/cnl2)

### DIFF
--- a/module/web/cnl_app.py
+++ b/module/web/cnl_app.py
@@ -67,11 +67,17 @@ def addcrypted():
     dlc_file.write(dlc)
     dlc_file.close()
 
+    pw = request.POST['passwords']
+
     try:
-        PYLOAD.addPackage(package, [dlc_path], 0)
+        pack = PYLOAD.addPackage(package, [dlc_path], 0)
     except:
         return HTTPError()
     else:
+        if pw:
+            pw = pw.decode("utf8", "ignore")
+            data = {"password": pw}
+            PYLOAD.setPackageData(pack, data)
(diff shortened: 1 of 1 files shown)


GammaC0de:
Thanks

GammaC0de:
Do you have some sample URL with password for testing?

ajayvarghese2000:
> Do you have some sample URL with password for testing?

No personal links however, at [JDownloader cnl2 standard](https://jdownloader.org/knowledge/wiki/glossary/cnl2) you will find 2 testing buttons for the 2 main endpoints for the click 'n' load links. `/flash/addcrypted2` and `/flash/add` the first endpoint is for links that are encrypted and the second for plain links.

(regardless of how the links are sent the password is send plainly anyway so the same code was used in both function)

You can use those two buttons and the pyload running locally or remotely to test the code, if you are using a remote instance you will need an extension/rule to redirect `http://127.0.0.1:9666` to the URL of your pyload instance.

PR #4089, opened 2022-02-04
[Main] Added "starting" downloads to "restart failed"

c0rrre:
Hi, 
I made a pull request cause I often got downloads stuck in status "starting" without ever downloading them.

### Describe the changes

I added status 7 (= starting) to the "restart failed" function.

Sometimes downloads stuck in status "starting" and wont ever start downloading. 
I changed it to restart those on "restart failed" function

### DIFF
--- a/src/pyload/core/database/file_database.py
+++ b/src/pyload/core/database/file_database.py
@@ -407,7 +407,7 @@ class FileDatabaseMethods:
 
     @style.queue
     def restart_failed(self):
-        self.c.execute("UPDATE links SET status=3,error='' WHERE status IN (6, 8, 9)")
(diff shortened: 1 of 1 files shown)


GammaC0de:
@c0rrre Thanks for your contribution but "starting" status cannot be considered as failed.

First, update your installation to the latest develop branch:
```
pip install --upgrade https://github.com/pyload/pyload/archive/refs/heads/develop.zip
```
and if the issue continues then we should investigate why this happens.

One Issue I can spot is that if pyload was stopped when a plugin is in status starting (and even "waiting") it will remain in this status even when pyload starts again.
This has to be fixed.

Issue #4186, opened 2022-08-15
Handle HTTP 503 for FileStore.to Plugin (Add auto retry or deparallelize downloads)


foorschtbar:
### Feature description
If you start too many downloads at FileStore at the same time (not parallel, just too short after each other), sometimes the server responses with a HTTP 503 error. It is possible to retry the download a few seconds later? In my test, 10-20s are enough. Or, it is possible to queue all downlods files and start them not together, but delays each download by 10-20s?

### Problem description
If you start a package with multiple files, some download fails with HTTP 503 Error. After retry with some delay, it works fine.

### Additional references

![grafik](https://user-images.githubusercontent.com/10727275/184691548-6b761d09-d78a-4617-8d3c-d50691e4a48d.png)

PR #4187, opened 2022-08-15
Add Bad Header (HTTP 503) Retry Feature

foorschtbar:
### Describe the changes

A Download will retried if they fail with HTTP 503. This happed on this hoster if you download to many file at the same time. 

Yes, see #4186


GammaC0de:
Please update the PR to the current develop branch.

foorschtbar:
Update done. But, unfortunately there is still a small problem. The "nice filename" that the hoster delivers (File1.zip) is no longer written on retry but the download ID is taken (4SUM6HQMM1.zip). I guess this is related to the previous HTTP 503, but I don't know how to solve it. Any idea?

GammaC0de:
Please increase plugin version.

HTTP 503 is most probably because of cloudflare.

PR #4200, opened 2022-10-01
Changes to the HotFolder plugin

msc94:
### Describe the changes

This PR includes two small changes to the HotFolder plugin:

1. Possibility to filter extensions in the HotFolder. This way one can add (for example) their Downloads directory and only filter out dlc files in there.
2. Add the ability to add packages to the collector and not only the queue

Nope


GammaC0de:
Thanks for you contribution but please make the PR against the `develop` branch.

PR #4243, opened 2023-01-25
Reenabled local plugins

StefftheEmperor2:
Local plugins did not work

reenabled the search in both locations


StefftheEmperor2:
should be fixed by #4248

PR #4245, opened 2023-01-26
added possibility to wait for something in interactive recaptcha solving

StefftheEmperor2:
### Describe the changes

I wrote a local plugin for some site .. the problem is, it is heavily loading contents with ajax and recaptcha does not work. i added the possibility in the recaptcha script to wait for someting.

i then use it like this:

```
response = self.captcha.challenge(key=sitekey, data=html, version='2js', script_params={"wait_for": 'iframe', 'wait_for_target': 'div#g-recaptcha > div > div'})
```


src/pyload/plugins/anticaptchas/ReCaptcha.py
```diff
@@ -42,57 +42,130 @@ class ReCaptcha(CaptchaService):
                                 "cb636da134c24633f2b2b38f56dfbb92"
 
     RECAPTCHA_INTERACTIVE_JS = """
```
StefftheEmperor2, line 44: TODO: Update the signature

GammaC0de:
Hello,
First, sorry for so late.

In order to check this, can you please send me the use case for this (you can do it by mail if you do not want to do it publicly)?

PR #4253, opened 2023-02-07
added request method

StefftheEmperor2:
### Describe the changes

Added argument "method" to http_request and added argument to calling methods

In my plugin I need to make a put request


GammaC0de:
What is the use of this put request, upload a file?

StefftheEmperor2:
It's just the login data, which is sent as JSON in the request body

Issue #3277, opened 2018-11-24
ExtractArchive Crashes on non existing files


protree:
ExtractArchive Crashes on non existing files.

**Steps to reproduce**: Download .rar file, let it extract. Delete the rar file while keeping its entry in queque. restart pyload. ExtractArchive will crash with following log messages and won't extract any archives until entry is removed from queque and pyload is restarted:

> pyload_1          | 24.11.2018 17:56:43 INFO      ADDON ExtractArchive: Check package: packageRenamed
> pyload_1          | Exception in thread Thread-8:
> pyload_1          | Traceback (most recent call last):
> pyload_1          |   File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
> pyload_1          |     self.run()
> pyload_1          |   File "/opt/pyload/module/PluginThread.py", line 459, in run
> pyload_1          |     self.f(*self.args, **self.kwargs)
> pyload_1          |   File "/opt/pyload/module/plugins/hooks/ExtractArchive.py", line 178, in extract_queued
> pyload_1          |     if self.extract(packages, thread):  # @NOTE: check only if all gone fine, no failed reporting for now
> pyload_1          |   File "/opt/pyload/module/plugins/hooks/ExtractArchive.py", line 368, in extract
> pyload_1          |     if len(os.listdir(pack_dl_folder)) == 0:
> pyload_1          | OSError: [Errno 2] No such file or directory: '/opt/pyload/Downloads/Download/packageRenamed/'

GammaC0de:
> Steps to reproduce: Download .rar file, let it extract. 
Delete the rar file while keeping its entry in queue. 

Have you deleted the .rar file before the extraction completed?

protree:
No, I deleted the file after extraction finished. It also happens when you enable „delete packages after successful extraction“ in extractarchive settings.

GammaC0de:
What is the package status after the process?

Debug log please..

protree:
Following Situation with debug log:

2 packages:
- _oldPackageThatIsInQuequeButWasDeletedAtSomePoint_: ExtractArchive crashes on this package on restart, unfortunally I don't have logs how this package was downloaded and extracted
[...]

PR #4281, opened 2023-04-23
Add SIGTERM handler for graceful shutdowns (e.g. when using Docker)

dulli:
### Describe the changes

Registers a handler for the `SIGTERM` signal that simply calls the `terminate()` function, analogous to the existing `SIGQUIT` handler. 

Using _Docker_ to handle the lifecycle of a _pyload-ng_ container, i.e. with `docker stop pyload-ng` or `docker restart pyload-ng`, resulted in corrupted states where on the next startup _pyload_ could not be started and an error message stating `pyLoad already running with pid 1` was shown.

This is due to _Docker_ using `kill -s SIGTERM 1` to request a graceful shutdown for the process at PID 1 when stopping a container. As this signal is currently not handled by _pyload-ng_, it waits for 10 seconds and then forcefully stops the process, which doesn't correctly remove the `pidfile`.

### Additional references

This problem was mentioned in #3277, but I don't know if it directly caused that issue.

### DIFF
--- a/src/pyload/core/__init__.py
+++ b/src/pyload/core/__init__.py
@@ -195,10 +195,9 @@ class Core:
         from .managers.addon_manager import AddonManager
         from .managers.captcha_manager import CaptchaManager
         from .managers.event_manager import EventManager
+        from .managers.file_manager import FileManager
         from .managers.plugin_manager import PluginManager
         from .managers.thread_manager import ThreadManager
-        from .managers.file_manager import FileManager
-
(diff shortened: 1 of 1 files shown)


GammaC0de:
Hi, please issue the PR against the develop brabch, thanks.

dulli:
Sorry, I got a little distracted earlier while creating the PR. Should be changed now!

GammaC0de:
👍

Issue #4282, opened 2023-04-23
Pyload Next doesn't use doesn't use Google drive plugin


TheHacker66:
### Description

Pyload Next doesn't use doesn't use Google drive plugin for this link:

https://drive.google.com/u/0/uc?id=1-_hNJvmoBVGYGXXUierSpCuvYLvYfx7g&export=download

### Debug log

<html><body>

1413 | 2023-04-23 21:59:04 | INFO | pyload | Added package deadspace containing 1 links
-- | -- | -- | -- | --
1414 | 2023-04-23 21:59:04 | INFO | pyload | Download starts: https://drive.google.com/u/0/uc?id=1-_hNJvmoBVGYGXXUierSpCuvYLvYfx7g&export=download
1415 | 2023-04-23 21:59:04 | WARNING | pyload | DOWNLOADER DefaultPlugin[3]: Plugin may be unstable
1416 | 2023-04-23 21:59:04 | INFO | pyload | DOWNLOADER DefaultPlugin[3]: Grabbing link info...
1417 | 2023-04-23 21:59:04 | INFO | pyload | DOWNLOADER DefaultPlugin[3]: Link name: uc
1418 | 2023-04-23 21:59:04 | INFO | pyload | DOWNLOADER DefaultPlugin[3]: Link size: N/D
1419 | 2023-04-23 21:59:04 | INFO | pyload | DOWNLOADER DefaultPlugin[3]: Link status: starting
1420 | 2023-04-23 21:59:04 | INFO | pyload | DOWNLOADER DefaultPlugin[3]: Processing url: https://drive.google.com/u/0/uc?id=1-_hNJvmoBVGYGXXUierSpCuvYLvYfx7g&export=download
1421 | 2023-04-23 21:59:05 | INFO | pyload | DOWNLOADER DefaultPlugin[3]: File saved
1422 | 2023-04-23 21:59:05 | WARNING | pyload | DOWNLOADER DefaultPlugin[3]: Check result: Html file \| Waiting 1 minute and retry
1423 | 2023-04-23 21:59:05 | INFO | pyload | DOWNLOADER DefaultPlugin[3]: Waiting 1 minute...
1424 | 2023-04-23 21:59:31 | INFO | pyload | Download aborted: uc

</body>
</html>

malkstar:
FYI this is because of the drive user selector. If you remove the `/u/0/` it will match the regex and perform as expected

TheHacker66:
Thanks for the instant fix!

[...]

PR #4285, opened 2023-04-24
Google drive multi-user regex fix

malkstar:
### Describe the changes

Google drive links can sometimes contain a user indicator if you are logged into multiple accounts in the browser where the link is generated.

#4282 

### Additional references

I'm not sure if this an issue which needs fixing, as it can be worked around by the user. However, supporting the URLs is not a massive change to the regex.

### DIFF
--- a/src/pyload/plugins/downloaders/GoogledriveCom.py
+++ b/src/pyload/plugins/downloaders/GoogledriveCom.py
@@ -20,7 +20,7 @@ from ..base.downloader import BaseDownloader
 class GoogledriveCom(BaseDownloader):
     __name__ = "GoogledriveCom"
     __type__ = "downloader"
-    __version__ = "0.34"
+    __version__ = "0.35"
     __status__ = "testing"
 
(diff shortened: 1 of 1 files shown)


src/pyload/plugins/downloaders/GoogledriveCom.py
```diff
@@ -18,7 +20,7 @@
 class GoogledriveCom(BaseDownloader):
     __name__ = "GoogledriveCom"
     __type__ = "downloader"
-    __version__ = "0.34"
+    __version__ = "0.35"
     __status__ = "testing"
 
     __pattern__ = r"https?://(?:www\.)?(?:drive|docs)\.google\.com/(?:file/d/|uc\?.*id=)(?P<ID>[-\w]+)"
```
malkstar, line 26:
```suggestion
    __pattern__ = r"https?://(?:www\.)?(?:drive|docs)\.google\.com/(?:u/[0-9]/)?(?:file/)?(?:u/[0-9]/)?(?:d/|uc\?.*id=)(?P<ID>[-\w]+)"
```

not sure how this was lost from the PR

PR #4361, opened 2023-08-24
fix missing tourbobit.com url since last commits

Web-Eye:
### Describe the changes

fix missing tourbobit.com url since last commits


GammaC0de:
As far as I can see, `tourbobit.com` was never included before.

PR #4370, opened 2023-09-15
Update a test expression to fix a logical short circuit

munahaf:
In file: ExtractArchive.py, method: `remove`, a logical expression uses the identity operator. A new object is created inside the identity check operation and then used for matching identity. Since this is a distinct, new object, it will not have identity an match with anything else. As a result, the identity check will have a logical short circuit and the program may have unintended behavior. 

In this case, the following binary operation
        
      queue is []

compares a newly created object with the identity operator which will always evaluate to False.

I suggested that the logical operation should be reviewed for correctness. 

Sponsorship and Support:

This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

### DIFF
--- a/src/pyload/plugins/accounts/FshareVn.py
+++ b/src/pyload/plugins/accounts/FshareVn.py
@@ -10,7 +10,7 @@ from ..base.account import BaseAccount
 class FshareVn(BaseAccount):
     __name__ = "FshareVn"
     __type__ = "account"
-    __version__ = "0.27"
+    __version__ = "0.28"
     __status__ = "testing"
 
     __description__ = """Fshare.vn account plugin"""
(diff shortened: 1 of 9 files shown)


GammaC0de:
Thanks for your contribution but please issue the PR against the `develop` branch.

PR #4385, opened 2023-11-01
[Turbobit.com] fixed premium link pattern

Web-Eye:
### Describe the changes

Fixed const "LINK_PREMIUM_PATTERN"

The affected only Premium users:
Turbobit.net falls back to free download for every link, it seems they changed the premium download site.


GammaC0de:
you cto check/an use https://www.debuggex.com/  to check/author regexs.

PR #4395, opened 2023-11-16
[UnRar] Allow extraction ignoring file attributes

mihawk90:
### Describe the changes
This adds a new (default off) option which if enabled passes `-ai` to `unrar`, which ignores file attributes stored in RAR archives.

This should help with cases where file permissions are set to odd values that may result in users not being able to modify extracted files. By ignoring attributes, permissions are handled by the underlying OS functions of creating a file.

I run pyload-ng on a NAS where file and directory permissions are set up using ACLs. When unpacking RAR archives I was greeted with permission errors when trying to move or rename files that were inside nested directories in the RAR archive. Some of the files also had odd permissions set by default. This should™ alleviate this issue entirely. And if not... well, I'm sure it'll help *someone*.

### Additional references

The same issue was discussed 9 years(!) ago on the NZBGet forums and this sollution seems to have helped the user:
https://forum.nzbget.net/viewtopic.php?t=1550
edit:
Tried it out from inside the Docker container's shell and can confirm this also solved my permissions issue with an archive.

### DIFF
--- a/src/pyload/plugins/extractors/UnRar.py
+++ b/src/pyload/plugins/extractors/UnRar.py
@@ -13,10 +13,13 @@ from pyload.plugins.helpers import renice
 class UnRar(BaseExtractor):
     __name__ = "UnRar"
     __type__ = "extractor"
-    __version__ = "1.47"
+    __version__ = "1.48"
     __status__ = "testing"
 
-    __config__ = [("ignore_warnings", "bool", "Ignore unrar warnings", False)]
+    __config__ = [
+        ("ignore_warnings", "bool", "Ignore unrar warnings", False),
+        ("ignore_file_attributes", "bool", "Ignore File Attributes", False)
+    ]
 
     __description__ = """RAR extractor plugin"""
     __license__ = "GPLv3"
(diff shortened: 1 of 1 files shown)


mihawk90:
@GammaC0de I just now noticed 5b8e90065193f9ebd9c5fecff8ae0c4f9cd76e8e after this PR.

Correct me if I'm wrong but I'm pretty sure this turns it on when the user turns it (or leaves it) off. The option is off by default:
https://github.com/pyload/pyload/blob/695bb70cd88608dc4fee18a6a7ecb66722ebfd8f/src/pyload/plugins/extractors/UnRar.py#L21

And with your commit it compares the value of the config with `False`
https://github.com/pyload/pyload/blob/695bb70cd88608dc4fee18a6a7ecb66722ebfd8f/src/pyload/plugins/extractors/UnRar.py#L280

Which would mean if the user turns it off the config variable is `False`, compares it with `False`, which returns `True` and therefore turns the option on, no? Or am I misunderstanding the second attribute here?

I'm fairly sure when I tested it, it worked as expected :thinking:

GammaC0de:
The second parameter for get() is optional and its the default value when the key does not exist. https://www.w3schools.com/python/ref_dictionary_get.asp

mihawk90:
Oohhh that makes sense then, sorry.

Issue #4401, opened 2023-11-28
[XDCC] UnicodeDecodeError on channel join with message containing non-ASCII chars


jonas-klein:
### Description
If a NOTICE containing non-ASCII characters (e.g. German umlauts like äöü) is sent to pyload while pyload is connected in a channel, a crash occurs.

`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe4 in position 96: invalid continuation byte`

I am not quite sure that it is a NOTICE. In any case, it is a message that is sent at the moment pyload has connected to a channel.

### Debug log
The following stack trace occurs
```
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/pyload/core/threads/download_thread.py", line 56, in run
    pyfile.plugin.preprocessing(self)
  File "/usr/lib/python3.10/site-packages/pyload/plugins/base/hoster.py", line 291, in preprocessing
    return self._process(*args, **kwargs)
  File "/usr/lib/python3.10/site-packages/pyload/plugins/base/downloader.py", line 103, in _process
    self.process(self.pyfile)
  File "/usr/lib/python3.10/site-packages/pyload/plugins/downloaders/XDCC.py", line 755, in process
    if not self.irc_client.join_channel(chan):
  File "/usr/lib/python3.10/site-packages/pyload/core/utils/struct/lock.py", line 22, in wrapped
    return func(self, *args, **kwargs)
  File "/usr/lib/python3.10/site-packages/pyload/plugins/downloaders/XDCC.py", line 246, in join_channel
    origin, command, args = self.get_irc_command()
  File "/usr/lib/python3.10/site-packages/pyload/core/utils/struct/lock.py", line 22, in wrapped
    return func(self, *args, **kwargs)
  File "/usr/lib/python3.10/site-packages/pyload/plugins/downloaders/XDCC.py", line 171, in get_irc_command
    line = self._get_response_line()
  File "/usr/lib/python3.10/site-packages/pyload/plugins/downloaders/XDCC.py", line 60, in _get_response_line
    self.receive_buffer += to_str(self.irc_sock.recv(1 << 10))
  File "/usr/lib/python3.10/site-packages/pyload/core/utils/convert.py", line 82, in to_str
    return obj.decode(encoding, errors)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe4 in position 96: invalid continuation byte
```

### Additional info
I could resolve this issue by adding `encoding="latin-1"` in XDCC.py

```
[...]

PR #4403, opened 2023-11-28
resolved utf8 decoding error

jonas-klein:
### Describe the changes

Just changed the default UTF-8 encoding to Latin-1.

https://github.com/pyload/pyload/issues/4401
UnicodeDecodeError on channel join with message containing non-ASCII chars

If the response line contains non-ASCII chars, line 60 was throwing an exception.
`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe4 in position 96: invalid continuation byte`


GammaC0de:
https://github.com/pyload/pyload/commit/40aa654d47594609f064d9955571900b5c32ab26

PR #4446, opened 2024-04-09
Replace dependency Js2py with dukpy

tr4nt0r:
### Describe the changes

Replace js2py dependency with dukpy

js2py is not compatible with Python 3.12 and prevents pyLoad from starting. There may be more incompatibilities with Python 3.12 but at least pyLoad now starts.


GammaC0de:
Replace dependency Js2py with dukpy for >= 3.12

PR #4480, opened 2024-06-07
[Crypter & Downloader] Added knigavuhe.org

EnergoStalin:
### Describe the changes

Add simple provider for https://knigavuhe.org so feeding url like this `https://knigavuhe.org/book/dopolnitelnye-istorii/` would download a book.

S in SKnigavuhe states for `Storage` cause i suspect s3 backend judging by subdomain. It serves sole purpose - rewriting name without touching url. Couldn't figure out better way to do it.

### Additional references

image

Maybe i will add proper naming templates to it and more metadata resolving from page but for now it's all what i need.

### DIFF
--- a/src/pyload/plugins/decrypters/Knigavuhe.py
+++ b/src/pyload/plugins/decrypters/Knigavuhe.py
@@ -1,7 +1,10 @@
 # -*- coding: utf-8 -*-
 
 from ..base.decrypter import BaseDecrypter
-import re, json, os
+
+import re
+import json
+import os
 
 
 class Knigavuhe(BaseDecrypter):
--- a/src/pyload/plugins/decrypters/Knigavuhe.py
+++ b/src/pyload/plugins/decrypters/Knigavuhe.py
@@ -3,6 +3,7 @@
 from ..base.decrypter import BaseDecrypter
 import re, json, os
 
+
 class Knigavuhe(BaseDecrypter):
     __name__ = "Knigavuhe"
     __type__ = "decrypter"
@@ -15,7 +16,7 @@ class Knigavuhe(BaseDecrypter):
 
     BOOK_PLAYER_DATA_PATTERN = r"new BookPlayer\(\d+,\s?(\[.+?\]),"
     BOOK_TITLE_PATTERN = r"book = ({.+});"
-    VOLUME_PATTERN = r'((?:\d+.?)\d+)\.<\/span>\s+<strong>{}'
+    VOLUME_PATTERN = r"((?:\d+.?)\d+)\.<\/span>\s+<strong>{}"
 
     def _get_book_title(self, html):
         text = re.search(self.BOOK_TITLE_PATTERN, html)
@@ -25,7 +26,7 @@ class Knigavuhe(BaseDecrypter):
 
         title = json.loads(text[1])["name"]
         volume = re.search(self.VOLUME_PATTERN.format(title), html)
-        if volume: # If single volume omit number in series
+        if volume:  # If single volume omit number in series
             title = f"{volume.group(1)}. {title}"
 
         return title
@@ -49,7 +50,10 @@ class Knigavuhe(BaseDecrypter):
         self.packages.append(
             (
                 title or pyfile.package().name,
-                [f"{x['url']}#{x['title']}{os.path.splitext(x['url'])[1]}" for x in data],
+                [
+                    f"{x['url']}#{x['title']}{os.path.splitext(x['url'])[1]}"
+                    for x in data
+                ],
                 title or pyfile.package().folder,
             )
         )
--- a/src/pyload/plugins/decrypters/Knigavuhe.py
+++ b/src/pyload/plugins/decrypters/Knigavuhe.py
@@ -1,0 +1,55 @@
+# -*- coding: utf-8 -*-
+
+from ..base.decrypter import BaseDecrypter
+import re, json, os
+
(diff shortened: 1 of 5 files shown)


src/pyload/plugins/decrypters/Knigavuhe.py
```diff
@@ -0,0 +1,62 @@
+# -*- coding: utf-8 -*-
+
+from ..base.decrypter import BaseDecrypter
+
+import re
+import json
+import os
+
+
+class Knigavuhe(BaseDecrypter):
```
GammaC0de, line 10: class name has to be called `KnigavuheOrg`

src/pyload/plugins/decrypters/Knigavuhe.py
```diff
@@ -0,0 +1,62 @@
...
+    __name__ = "Knigavuhe"
```
GammaC0de, line 11: also `__name__ ` has to be called `KnigavuheOrg`

src/pyload/plugins/decrypters/Knigavuhe.py
```diff
@@ -0,0 +1,62 @@
...
+    __type__ = "decrypter"
+    __version__ = "0.1"
```
GammaC0de, line 13: initial version should be set to "0.01"

src/pyload/plugins/downloaders/SKnigavuhe.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+
+from ..base.simple_downloader import SimpleDownloader
+
+
+class SKnigavuhe(SimpleDownloader):
+    __name__ = "SKnigavuhe"
```
GammaC0de, line 7: class name and the python filename has to be called `KnigavuheOrg` and `KnigavuheOrg.py`  accordingly
`__name__` has to be called `KnigavuheOrg`

src/pyload/plugins/downloaders/SKnigavuhe.py
```diff
@@ -0,0 +1,21 @@
...
+    __type__ = "downloader"
+    __version__ = "0.1"
+    __status__ = "testing"
```
GammaC0de, line 10: initial version should be set to "0.01"

src/pyload/plugins/downloaders/SKnigavuhe.py
```diff
@@ -0,0 +1,21 @@
...
+    __pattern__ = (
```
GammaC0de, line 11: `__pattern__ ` must not be surrounded by round brackets

EnergoStalin: It does automatically by black due to line length > 88(89 to be exact). I guess linter will fail if i change that.

GammaC0de:
Sorry but one more thing: please issue the PR against the `develop` branch

EnergoStalin:
Also it suddenly stopped working now with `Loaded URL exceeded limit (2000000)` exception investigating...

GammaC0de:
>Also it suddenly stopped working now with Loaded URL exceeded limit (2000000) exception investigating...

Probably a solution would be something like:
https://github.com/pyload/pyload/blob/9376854e794f4c3eaab8fd3cb9025994d359c2c4/src/pyload/plugins/decrypters/MegaCoNzFolder.py#L34-L44

EnergoStalin:
Don't think it should be that big not bigger 2GB for sure. Will debug through mitmproxy to see if it actually should be this big.

EnergoStalin:
Ok it's broken because i renamed Hoster part. When adding S prefix again it works like before. Probably i guessed first time right you cant name it same way expecting to work.

If you know way to set filename in `Decrypter` we can get rid of hoster completely.

EnergoStalin:
> > Also it suddenly stopped working now with Loaded URL exceeded limit (2000000) exception investigating...
> 
> Probably a solution would be something like:
> 
> https://github.com/pyload/pyload/blob/9376854e794f4c3eaab8fd3cb9025994d359c2c4/src/pyload/plugins/decrypters/MegaCoNzFolder.py#L34-L44

This is fixing the issue but then it fails unable to resolve BookPlayer data in decrypter.

EnergoStalin:
It's not working on 0.5.0 version in current state(Downloader renamed without S prefix). I already described problem along with solutions i can see [above](https://github.com/pyload/pyload/pull/4480#issuecomment-2156227890). Of course we could delegate naming to default Downloader but then you need rename files manually for sane names.

EnergoStalin:
I blindly suspect the culprit to be recursion in case with decrypter and downloader being named the same.

Besides that it's the only plugin that try to repeat it's name across given types.

```sh
ls src/pyload/plugins/decrypters > /tmp/dec.txt
ls src/pyload/plugins/downloaders > /tmp/down.txt

comm -12 /tmp/dec.txt /tmp/down.txt

# __init__.py
# KnigavuheOrg.py
# __pycache__
```

GammaC0de:
> I blindly suspect the culprit to be recursion in case with decrypter and downloader being named the same.

You are right, 
as a rule of a thumb, if there are both a downloader and a decrypter, the decrypter is usually called HosternameFolder.py.
So in your case it souldd be called `KnigavuheOrgFolder.py` and the decrypter calss sould be called `KnigavuheOrgFolder`.

EnergoStalin:
It's pretty much ready to merge if you don't have more naming questions.

PR #4478, opened 2024-06-01
fetch missing SSL certificates

milahu:
fix #4469

based on https://github.com/danilobellini/aia/pull/3

webui seems broken
i can login, but after 15...20 seconds, i get logged out
no errors in debug log

maybe because im running 2 pyload in parallel
im starting pyload 2 with

```
pyload --debug --userdir ~/.pyload2 --tempdir /tmp/pyLoad2 \
  --pidfile /tmp/pyLoad2/pyload.pid
```

PR #4486, opened 2024-06-16
dynamic proxy per download

### LINKED ISSUE #3899: Per link (download) proxy config
I am not sure how easy is to implement such a feature, but for sure it is a great one, I think it is necessary to make the download (any link) has specific configuration for proxy, and weither to use it or not, and maybe adding multiple proxies??

the current situation which is a proxy for pyload in general doesn't suit my need, as some services blocks VPNs and I need to use for example TOR for some links, and for others no proxy.

milahu:
fix #3899

based on #4478 but not required here

work in progress, including some other changes like hot-reloading

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -83,6 +83,12 @@ install_requires =
     dukPy>=0.3.1;python_version>="3.12"
     pycurl~=7.43
     certifi
+
+    # aia # TODO use pypi relase instead of src/pyload/core/network/http/aia.py
+    # extra deps for aia:
+    pyopenssl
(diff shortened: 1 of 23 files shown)


Genio40921 (approved):

milahu:
@Genio40921 the proxy feature is not working, so congrats, you approved a pile of garbage

i dont know why pycurl over tor is failing
now i got a debrid account, so i dont need this feature any more

some commits may be useful, but i dont care to get them merged
pyload would need a complete rewrite for proper hot-reloading

closing

PR #4495, opened 2024-07-12
Update FastshareCz.py - also match fastshare.cloud

andree182:
Make the pattern more generic, as it works on both...

### Describe the changes

This just extends the respective matching pattern, it seems to work just fine afterwards.

no.


GammaC0de:
Please issue the PR against the `develop` branch.

vuolter:
👍

PR #4518, opened 2024-10-11
add SerienfansOrg

milahu:
scrape `filecrypt.cc` urls from `serienfans.org`

based on my [serienfans-scraper](https://github.com/milahu/serienfans-scraper)

status: broken, because `FilecryptCc` is still broken
and probably fails on cutcaptcha, which requires `2captcha.com` solver


GammaC0de:
First of all,
This site serienfans is not legal an so is cannot be added.
Secondly you put file that are totally unrelated to the PR such as shell.nix and much more 😕

PR #4543, opened 2025-02-26
Add serienfans org

milahu450:
rebased version of #4518 

happy closing, moving on : )


AlexanderS:
It seems clear, that this cannot be merged as is. This is a mix of a lot of different changes and contains extra files (`*.trash`?!).

Issue #4414, opened 2023-12-30
Remote Control


michnixweiss:
### Feature description
I am missing the remote section in settings.

### Problem description
In pyload 0.4.20 I can configure and use remote (pyload apk) on port 7227. This is not implemented / ported to 0.5.0

GammaC0de:
@vuolter Do you have an new idea how to integrate the android app again or should we re-add the thrift backend?

michnixweiss:
Thanks @vuolter for your answer, but my python skills are limited, so no suggestions on how to integrate from my side.
I just want to use pyload next and miss this feature.

vuolter:
@GammaC0de I have no plans for the android app right now.
Actually, if we improve the webui for mobile and tablet screens, I'd like more to keep a single lite (web) app than supporting android or ios natives.

michnixweiss:
Is there going to be a decision/solution?
I would like to further use the app. It is more usable on mobile than the web.

THX

GammaC0de:
I'm too in favor of using native app, I just need to find  spare time for it.

vuolter:
> If the is going a SPA UI that uses Angular or React, the API should be flexible enough to allow the use of native apps too. 

👍

michnixweiss:
Is there any progress?

vuolter:
Still nothing from my side... @GammaC0de if you would make a try with thrift I guess it's the quick way to restore android/remote api capability in the meantime.

michnixweiss:
@GammaC0de are there any news or something for testing?

GammaC0de:
after looking at the code for remote control I that the whole API access needs a complete rework and modernized using REST and OpenAPI for example so the maintainability would be better.

I tried porting the Thrift but it seems a lot of work too.
Still not gave up though..

vuolter:
@GammaC0de thrift support was removed when the codebase was rewritten to work on Python 3 to reduce complexity and throw away oldish deps. But all the interfaces exposed by thrift was actually a 1:1 copy of the API available now through flask, so another way to restore the official android client could be update the android app to connect to pyLoad via HTTP instead of the (thrift) socket.

michnixweiss:
Is there any progress?

vuolter:
Still no progress, sry.

michnixweiss:
Groundhog Day ;-)
Any news?

michnixweiss:
It's a year now, is this going to be implemented?

GammaC0de:
[...]

PR #4560, opened 2025-03-24
Introduce Pydantic data models and API type annotations

Sab44:
### Describe the changes

As outlined in [my comment](https://github.com/pyload/pyload/issues/4414#issuecomment-2742702273) on [#4414](https://github.com/pyload/pyload/issues/4414) this is the first step to enable automatic API documentation.

- All API methods are type annotated, both parameters and return types
- All non-primitive data types are moved to Pydantic models for automatic validation

So far, everything seems to work just fine.

With this, we can parse the existing methods into an OpenAPI specificaton, will submit this in a follow-up PR. I have a working draft, but it needs some polishing.

I've tried to encapsulate all changes to logical commits to make the review easier. In any case, I'd say this should probably stay on the develop branch for a couple of weeks before merging into main to make sure it doesn't cause any issues.

### Additional references

Related issue #4414

### DIFF
--- a/src/pyload/core/api/__init__.py
+++ b/src/pyload/core/api/__init__.py
@@ -393,7 +393,7 @@ class Api:
     @permission(Perms.LIST)
     def status_downloads(self):
         """
-        Status off all currently running downloads.
+        Status of all currently running downloads.
 
         :return: list of `DownloadStatus`
         """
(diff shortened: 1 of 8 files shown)


src/pyload/core/api/__init__.py
```diff
@@ -1215,15 +1236,15 @@ def get_account_types(self):
 
     @legacy("updateAccount")
     @permission(Perms.ACCOUNTS)
-    def update_account(self, plugin, account, password=None, options={}):
+    def update_account(self, plugin: str, account: str, password: str = None, options: dict[str: Any] = {}) -> None:
```
Sab44, line 1239: I understand the code analysis is throwing a warning here, but I think for now we should keep the default arguments as they were before.

src/pyload/webui/app/blueprints/json_blueprint.py
```diff
@@ -31,46 +31,27 @@ def links():
...
-            package["links"] = []
-            for file in api.get_package_files(package["id"]):
-                package["links"].append(api.get_file_info(file))
-
-        return jsonify(data)
-
-    except Exception:
-        return jsonify(False), 500
-
-
```
Sab44, line 73: I deleted this function because I don't think it actually worked.

it calls `get_queue()` which returns `PackageData` and then tries to access it `id` - but `PackageData` has no `id` property/key.
Even if we assume that `pid` was meant here, then it will call `api.get_package_files()` which doesn't exist.

So something was very broken here. If you think we should keep it then please let me know how to fix this method to actually work as expected.

Sab44:
Closed, the change here are part of the follow-up PR that introduces the OpenAPI spec:
https://github.com/pyload/pyload/pull/4580

PR #4580, opened 2025-05-27
Add OpenAPI 3.1 compliant spec generator

Sab44:
### Describe the changes

**Motivation:** 
Migrate the no-longer functional pyLoad Android app from Thrift to the REST API currently in use in pyLoad.
Android PR: https://github.com/pyload/pyload-android/pull/43

**Summary:**
- create an OpenAPI spec generator that parses the REST API implementation and generates an API specification out of it
- keep changes to existing code and API as minimal as possible
- keep API backwards compatible to not break any existing clients
- generate API client automatically via the OpenAPI generator
- the OpenAPI specification will be exposed on 2 endpoints:
  - `/api/docs` for the Swagger UI representation
  - `/api/openapi.json` for the raw JSON specification

**Details:**
- all API functions must have their parameters and return values annotated
- in case of custom model types these were replaced by [Pydantic](https://github.com/pydantic/pydantic) models as they have in-built OpenAPI conformance
- docstring of API methods will be parsed for method, parameter and return value docs
- to conform with OpenAPI standards, the following logic is used to determine the appropriate REST method:
  - functions requiring no parameters will use a GET method
  - functions with primitive parameters will use a POST method with query params
  - functions with non-primitive parameters (e.g. arrays, dicts, pydantic models) will use a POST method with json request body
  - file uploads will use a POST method with multipart request body
  - this logic is derived from how [FastAPI handles this](https://fastapi.tiangolo.com/tutorial/body/#request-body-path-query-parameters)
- login via session cookie is hardcoded in the specification
  - you can login via Swagger UI and then execute API functions on Swagger UI directly
  - you can logout via Swagger UI as well
  - alternatively, you can copy a valid session cookie value into the `Authorize` button at the top
- a copy of the `openapi.json` is checked-in at `pyload/openapi-generator`
  - pyLoad can be run with `pyload --generate-api-spec` to update this json
  - added a test to the CI to make sure this copy is always up-to-date to make sure changes to the API do not break the specification
- end-of-life Python versions have been removed, the min version is now 3.9 (required by pydantic)
- to generate code for clients, use the [OpenAPI generator](https://github.com/OpenAPITools/openapi-generator)
  - instructions on how to generate the Android client are added to the README
  - of course any other client could also be generated, e.g. for a Python client library

I've tried to encapsulate all changes to logical commits to make the review easier. In any case, I'd say this should probably stay on the develop branch for a couple of weeks before merging into main to make sure it doesn't cause any issues.

@GammaC0de this was more work than I expected, but I got there eventually. Let me know what you think.

https://github.com/pyload/pyload/issues/4414

### Additional references
pyload_openapi
pyload_android


Sab44:
Glad to hear 🙂 

Of course, I'll keep testing it and if anything comes up post it here, but as you said the changes to the existing code are relatively minor so hopefully there are no major issues.

Sab44:
Updated the OpenAPI JSON to reflect the changes to `DownloadInfo`'s `wait_until` change.
The stored JSON is considered a source of truth to make sure any changes to the API do not break the API specification - if the CI had run, there's a check that compares the stored JSON to the generated one and fails if they are different.

Also re-generated the Android client to include the changes, see Android PR. Although the changes are not relevant for the app's functionality at the moment.

Btw good catch with the Captcha, I only tested with my Premium account so I missed that.

Malumatra:
- [ ] - > F for  ooo fFg- fa t``  of` ff` an ff

PR #4614, opened 2025-08-25
fix: Use of a broken or weak cryptographic algorithm

opsysdebug:
**General approach:**  
Replace uses of AES in ECB mode with a stronger, secure block mode, such as CBC (Cipher Block Chaining), with a random IV for each encryption. For decryption, the IV must be provided alongside the ciphertext (typically prepended), as it will be required to decrypt.

**Detailed step-by-step:**  
- Refactor the `ecb_encrypt` function to:
  - Use AES-CBC mode instead of AES-ECB.
  - Generate a random 16-byte IV for each encryption.
  - Prepend the IV to the ciphertext (so it can be used in decryption).
- Refactor the `ecb_decrypt` function to:
  - Read the first 16 bytes of its input as the IV.
  - Use the same AES-CBC mode to decrypt the remainder of the data.
- Add an import for `os` (for secure random IV generation) if not already imported above.
- Ensure `MegaCrypto.a32_to_bytes(key)` remains unmodified, to preserve key processing.
- Update calls within `MegaCrypto` accordingly (only within lines shown).
- Optionally update docstrings/comments to clarify the new IV handling, if present.

**Which lines to change:**  
- Lines 104–120 (the `ecb_decrypt` and `ecb_encrypt` methods in the shown code).

**Imports:**  
- `os` is already imported (line 5), so no further imports required.

---

Fix the use of the insecure ECB cipher mode, replace the use of `Cipher(..., modes.ECB(), ...)` with a more secure and appropriate approach for deriving or processing the IV, such as using a secure key derivation function (KDF) or a random IV. In this case, since the code is generating an IV with ECB mode (to use with CFB), it is likely following an old or non-standard derivation process.

The best solution is to generate a cryptographically secure random IV using `os.urandom()` of the correct block size (16 bytes for AES). This IV should then be used in the call to `Cipher(..., modes.CFB(iv), ...)`. Any dependence on the legacy ECB-processing step should be reviewed and, if backward compatibility must be preserved, it should be carefully documented; otherwise, the random IV approach is preferred.

**Changes:**
- In `src/pyload/plugins/containers/RSDF.py`, within the `decrypt()` method:
  - Replace the lines generating and processing the IV using ECB mode (`lines 49-55`) with code that generates a secure random IV of 16 bytes (`os.urandom(16)`).
- Update all relevant uses and flows to use the new IV.
- Document the change to indicate a secure IV derivation is now being used.

#### References
NIST, FIPS 140 Annex a: [Approved Security Functions](http://csrc.nist.gov/publications/fips/fips140-2/fips1402annexa.pdf)
NIST, SP 800-131A: [Transitions: Recommendation for Transitioning the Use of Cryptographic Algorithms and Key Lengths](http://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-131Ar1.pdf)
OWASP: [Rule - Use strong approved cryptographic algorithms](https://cheatsheetseries.owasp.org/cheatsheets/Cryptographic_Storage_Cheat_Sheet.html#rule---use-strong-approved-authenticated-encryption)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -9,7 +9,7 @@
     "python.pythonPath": ".venv/bin/python3.9",
     "[python]": {
         "editor.codeActionsOnSave": {
-            "source.organizeImports": true
+            "source.organizeImports": "explicit"
         }
     }
 }
--- a/src/pyload/plugins/containers/RSDF.py
+++ b/src/pyload/plugins/containers/RSDF.py
@@ -46,12 +46,12 @@ class RSDF(BaseContainer):
 
         backend = default_backend()
 
-        ecb = Cipher(algorithms.AES(KEY), modes.ECB(), backend=backend).encryptor()
-        iv = ecb.update(IV) + ecb.finalize()
+        # Use a securely generated random IV (16 bytes for AES)
+        iv = os.urandom(16)
 
-        cipher = Cipher(algorithms.AES(KEY), modes.CFB(iv), backend=backend)
-        encryptor = cipher.encryptor()
-        iv = encryptor.update(IV) + encryptor.finalize()
+        # Prepare cipher with secure IV
+        # No need to reprocess IV as before
+
 
         try:
             fs_filename = os.fsdecode(pyfile.url)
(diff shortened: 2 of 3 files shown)


GammaC0de:
RSDF and CCF  containers are deprecated https://jdownloader.org/knowledge/wiki/linkprotection/container/dlc-rsdf-and-ccf
they probably can be removed.

As for MegaCoNz.py, is this tested? are you sure there will be no issues downloading from  MegaCoNz?

PR #4633, opened 2025-10-15
Fix CI OpenAPI specification check

Sab44:
### Describe the changes

The current OpenAPI specification check on the pipeline doesn't really check anything, as the `--dry-run` parameter will prevent the json from being created and checked for changes in the next step.

What is the purpose of this check?

When someone makes changes to pyLoad API, it could potentially break the specification (I don't mean a breaking change, I mean that the resulting specification is invalid and cannot be used to generate client code).
The pipeline should make sure, that the local copy of the `openapi.json` is always up-to-date. If a mismatch is detected, that means there were changes made to the API that are not reflected in the specification. The developer should then re-create the local json file and verify that it changed as expected and is still valid.

What this PR does:
* Remove the api-spec test from the tox config
  * as tox is running inside the `.tox` workdir which is gitignored, any changes to the json will not be detected anyways
* Adapt the GitHub CI check to install pyload as editable install so the path to the json is reliable
* Updated the README with more information regarding the local api spec copy, as well as general development setup instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,23 @@ Invoke `build_locale` before building the package (eg. `bdist_wheel`).
 
 ## Development
 
+### Set up your development environment
+
+* Clone the repository
+* Recommended: create a virtual environment for pyLoad
+```
+python3 -m venv .venv
+source .venv/bin/activate
+```
+* Install pyLoad as editable install
+```
+pip install -e .
+```
+* Run pyLoad in debug mode
+```
+pyload -d
+```
+
 ### API specification
 
 pyLoad provides an OpenAPI specification for its REST API, visible via Swagger UI under the endpoint


GammaC0de:
>What is the purpose of this check?

Why I introduced the dry-run spec generator?
The `Test OpenAPI spec generator` job was failing because of path issues (see https://github.com/pyload/pyload/actions/runs/17221062531/job/48856358864) but the workflow was successful because `continue-on-error` was set, it was not noticeable and i found out by mistake.
I looked at the failing log  and tried to deduct what it was trying to achieve.
my conclusion it was testing if the spec generation is successful or not and so, i thought that writing to a file is not needed hence the dry-run.

PR #4636, opened 2025-10-30
Fix: New AlldebridCom API Path

technoo10201:
Fix for AlldebridCom API Path, according to the documentation:

This API version is namespaced as v4, as such all endpoint start with /v4/ or /v4.X/, such like http://api.alldebrid.com/v4/ping or http://api.alldebrid.com/v4.1/ping.

This API v4 should be the final version regarding general response format and errors (hopefully).


technoo10201:
Kind up @GammaC0de

GammaC0de:
Have you tested it with the new base URL? is it working flawlessly?

PR #4643, opened 2025-11-12
Basic auth in openapi spec

Sab44:
### Describe the changes

Not sure if intentional, but the `/api/login` endpoint can no longer be called with the current changes on `develop`.
The response will always be `{"error": "CSRF token is invalid"}`. Or am I missing something?

Anyway, with the support of http basic auth, we can switch to that instead which is an improvement in every way possible.

* adapted the openapi generator and recreated resulting openapi specification
* fixed some bugs with JSON schemas for optional properties

### Additional references

https://swagger.io/docs/specification/v3_0/authentication/basic-authentication/

### DIFF
--- a/openapi-generator/openapi.json
+++ b/openapi-generator/openapi.json
@@ -1,72 +1,16 @@
 {
   "info": {
     "title": "pyLoad API Documentation - OpenAPI",
-    "version": "1.0.0"
+    "version": "1.1.0"
   },
   "openapi": "3.1.1",
   "tags": [
-    {
-      "name": "pyLoad Authentication",
-      "description": ""
-    },
     {
       "name": "pyLoad REST",
       "description": ""
     }
   ],
   "paths": {
-    "/api/login": {
-      "post": {
-        "security": [],
-        "summary": "Login into pyLoad, this must be called when using rpc before any methods can be used.",
-        "tags": [
-          "pyLoad Authentication"
-        ],
-        "requestBody": {
-          "required": true,
-          "content": {
-            "application/x-www-form-urlencoded": {
-              "schema": {
-                "type": "object",
-                "properties": {
-                  "username": {
-                    "type": "string",
-                    "default": "pyload"
-                  },
-                  "password": {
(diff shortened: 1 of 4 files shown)


GammaC0de:
> Maybe the /api/login endpoint could return some kind of deprecation message just to make it transparent that it is no longer usable

You are absolutely  right.

milahu:
> Session based authentication is a target for [CSRF](https://en.wikipedia.org/wiki/Cross-site_request_forgery) attacks, so CSRF protection was added.

for API login with username and password
i am abusing the non-API login (`GET /login`) to get a `csrf_token`
and to submit the login form (`POST /login`) to get a session cookie
https://github.com/milahu/pyload-qt/commit/9fa7ccc12a706ea455267b600f3e4f4f8ad39ff0

i have to send `csrf_token` with every API request
to fix `CSRF Error: The CSRF token is missing.`
from `pyload/src/pyload/webui/app/handlers.py`

https://github.com/pyload/pyload/blob/3c5e1c667c17ae2cf0550408fe3a1543a4b0b068/src/pyload/webui/app/handlers.py#L28

and to fix `{"error": "Invalid API credentials"}`
from `pyload/src/pyload/webui/app/helpers.py`

https://github.com/pyload/pyload/blob/3c5e1c667c17ae2cf0550408fe3a1543a4b0b068/src/pyload/webui/app/helpers.py#L410

i had to remove `kwargs["csrf_token"]`
before passing `kwargs` to the API method
https://github.com/milahu/pyload/commit/655b5462a9a6347a56595091fcbb99d3a826a975
(please merge?)

to fix `TypeError: Api.get_config() got an unexpected keyword argument 'csrf_token'`
from `pyload/src/pyload/webui/app/blueprints/api_blueprint.py`
https://github.com/pyload/pyload/blob/3c5e1c667c17ae2cf0550408fe3a1543a4b0b068/src/pyload/webui/app/blueprints/api_blueprint.py#L76-L79

maybe i should use Basic Auth to access the API
but well... this works too : P

PR #4671, opened 2025-12-27
[OneFichierCom] Add DL_LIMIT_PATTERN

zep6yr:
### Describe the changes

Add wait pattern.

After a download finishes, the wait pattern changes, causing pyload to fail in recognizing the new pattern. This results in a temporary offline error and the next file being attempted instead of waiting for the specified time.


GammaC0de:
> Should I change the icon to \u23f3?

No, its just was quite a shock that it was used and its not a typo..

PR #4712, opened 2026-03-22
Fix API key error when sending requests

Sab44:
### Describe the changes

I have tested the intended flow:

1. Create API key
2. Send request with API key
```
curl -X 'GET' \
  'http://localhost:8000/api/status_server' \
  -H 'accept: application/json' \
  -H 'Authorization: Bearer pl_xxxxx'
```

Got an error:
```
  File ".../pyload/src/pyload/webui/app/helpers.py", line 265, in decorated_function
    timestamp = int(user_data["last_used"] * 1000)
                    ~~~~~~~~~^^^^^^^^^^^^^
KeyError: 'last_used'
```

It is fixed with this PR.

It also adds 2 more commits:

1. Revert to using the `flask.request.authorization` object for code readability
2. Remove clear text logging of the full api key

API keys not working at all.

### DIFF
--- a/openapi-generator/openapi.json
+++ b/openapi-generator/openapi.json
@@ -3443,9 +3443,9 @@
     },
     "securitySchemes": {
       "ApiKeyAuth": {
-        "type": "http",
-        "scheme": "bearer",
-        "bearerFormat": "custom"
+        "type": "apiKey",
+        "in": "header",
+        "name": "X-API-Key"
       }
     }
   },
--- a/src/pyload/webui/app/api_docs/openapi_specification_generator.py
+++ b/src/pyload/webui/app/api_docs/openapi_specification_generator.py
@@ -55,9 +55,9 @@ class OpenAPISpecificationGenerator:
                 "schemas": {},
                 "securitySchemes": {
                     "ApiKeyAuth": {
-                        "type": "http",
-                        "scheme": "bearer",
-                        "bearerFormat": "custom",
+                        "type": "apiKey",
+                        "in": "header",
+                        "name": "X-API-Key",
                     }
                 }
             },
--- a/src/pyload/webui/app/helpers.py
+++ b/src/pyload/webui/app/helpers.py
@@ -261,7 +261,7 @@ def apikey_auth(func):
                     user_data = api.pyload.db.get_all_user_data().get(user_id)
                     if user_data:
                         now = int(time.time() * 1000)
-                        timestamp = int(user_data["last_used"] * 1000)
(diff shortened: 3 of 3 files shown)


Sab44:
One more question: the intention is to add tokens in a way that can not be retrieved via a login call or similar. So an app user would have to create an API key, and copy-paste it into the app, making usernames & passwords obsolete (for API access), correct?

Why I'm asking is because the current design had me a bit confused. For API keys in that sense I did not expect them to be passed as bearer tokens but rather in a `X-API-Key` header.

https://swagger.io/docs/specification/v3_0/authentication/api-keys/
https://swagger.io/docs/specification/v3_0/authentication/bearer-authentication/

https://blog.postman.com/what-is-an-api-key/
https://blog.postman.com/what-is-a-bearer-token/

With a bearer token I expected this flow:

> 1. Client authenticates with credentials (username/password or OAuth flow).
> 
> 2. Server generates and returns a bearer token.

GammaC0de:
> One more question: the intention is to add tokens in a way that can not be retrieved via a login call or similar. So an app user would have to create an API key, and copy-paste it into the app, making usernames & passwords obsolete (for API access), correct?

Yes.

> for API keys in that sense I did not expect them to be passed as bearer tokens but rather in a X-API-Key header.

Well, it looked better to me but feel free to change to `X-API-Key` header

> Remove clear text logging of the full API key

Since it is a bad authorization it does not really matter since the key is invalid, I thought it wold be helpfull

Sab44:
> Well, it looked better to me but feel free to change to X-API-Key header

It's all good, just wanted to clarify.

> Since it is a bad authorization it does not really matter since the key is invalid, I thought it wold be helpfull

True, but you know, users will paste their password in there and suddenly you have credentials being logged.

In any case, feel free to change / refactor or just pick whatever you feel makes sense from this PR and discard the rest.

UserX404:
Just a question and my 2 cents.
What if doing the handshake/key exchange the other way?
As pyload is likely running on a server, desktop browser etc and the app on the phone, why not scaning the token as QR-code or optionaly copy a code from pyload GUI to android app?

GammaC0de:
I do agree with you and I've already thought about it but the would require to add another dependency to the project, so this is why I'm hesitating, I might do so eventually

GammaC0de:
@UserX404 https://github.com/pyload/pyload/commit/8fa0049fb7931616eb2053cd457d219f3ba345ce

UserX404:
@GammaC0de 
Just tested and it's working
related discussion: https://github.com/pyload/pyload-android/discussions/59

PR #4713, opened 2026-03-25
Fix first use of api key where last_used is not set and causes a crash

Sab44:
### Describe the changes

The very first time an API key is used, i.e. the first request that is made with it will cause an exception:
```
  File "../pyload/.venv/lib/python3.14/site-packages/flask/app.py", line 917, in full_dispatch_request
    rv = self.dispatch_request()
  File "../pyload/.venv/lib/python3.14/site-packages/flask/app.py", line 902, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "../pyload/src/pyload/webui/app/helpers.py", line 297, in decorated_function
    last_used = int(key_info["data"]["last_used"] * 1000)
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
TypeError: unsupported operand type(s) for *: 'NoneType' and 'int'
```
The reason for this is that `last_used` is not present in the database yet, it is only set in `src/pyload/core/api/__init__.py` **after** `key_data` has been retrieved. So the `key_info` at this point will not contain a value. This is fixed with this PR.

Also the conversion to `int` and multiplication by 1000 is already done when `last_used` is set / updated so I don't think it's correct to do it again when retrieving the value.

### DIFF
--- a/src/pyload/webui/app/helpers.py
+++ b/src/pyload/webui/app/helpers.py
@@ -293,8 +293,6 @@ def apikey_auth(func):
                 key_name = key_info["data"]["name"]
                 user_data = api.pyload.db.get_all_user_data().get(user_id)
                 if user_data:
-                    now = int(time.time() * 1000)
-                    last_used = int(key_info["data"]["last_used"] * 1000)
                     user_info = {
                         "id": user_id,
                         "name": user_data["name"],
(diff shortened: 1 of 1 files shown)


GammaC0de:
Thanks for reporting, this is better solution.
Old API keys have to be deleted.

Sab44:
@GammaC0de it is not crashing any longer now, but it is also not working as intended.

When `last_used` is retrieved in `helpers.py` it is multiplied by 1000:
https://github.com/pyload/pyload/blob/932a7cc323b6b06293844ab7fc542782e58814ca/src/pyload/webui/app/helpers.py#L297

When it is updated in `apikey_database.py` the value in the db is already multiplied by 1000:
https://github.com/pyload/pyload/blob/932a7cc323b6b06293844ab7fc542782e58814ca/src/pyload/core/database/apikey_database.py#L132

So it is multiplied twice by 1000 and therefore the condition `if now >= last_used + 3_600_000:` is never true.

Issue #4729, opened 2026-04-14
ddownload account broken in latest and dev branch


engels0n:
### Description

Hi,
when restarting pyload with newest version, ddownload account seems to be broken:

### Debug log

```
[2026-04-14 09:59:35]  INFO                pyload  ACCOUNT DdownloadCom: Grabbing account info for user `eng****`...
[2026-04-14 09:59:35]  DEBUG               pyload  ACCOUNT DdownloadCom: LOAD URL http://ddownload.com | get={'op': 'my_account'} | post=None | referrer=True | cookies=True | just_header=False | decode=True | multipart=False | redirect=True | req=None
[2026-04-14 09:59:36]  DEBUG               pyload  ACCOUNT DdownloadCom: VALID UNTIL PATTERN not found
[2026-04-14 09:59:36]  DEBUG               pyload  ACCOUNT DdownloadCom: Size: 200000 | Unit: MB
[2026-04-14 09:59:36]  DEBUG               pyload  ACCOUNT DdownloadCom: LEECH TRAFFIC PATTERN not found
[2026-04-14 09:59:36]  DEBUG               pyload  ACCOUNT DdownloadCom: Account info for user `eng****`: {'login': {'password': '**********', 'stats': [1, 0], 'timestamp': 1776153575.1942062, 'valid': True}, 'data': {'login': 'eng****', 'options': {'limit_dl': ['0']}, 'plugin': <Account DdownloadCom>, 'premium': False, 'trafficleft': 214748364800, 'type': 'DdownloadCom', 'validuntil': None, 'leechtraffic': None}}
```

### Additional references

<img width="1210" height="198" alt="Image" src="https://github.com/user-attachments/assets/0fe3c567-b890-4b8b-8568-8533dbdf5aab" />

Evironment:
- OS: Debian 13
- Docker Images:
image: lscr.io/linuxserver/pyload-ng:latest
image: lscr.io/linuxserver/pyload-ng:version-0.5.0b3.dev96
image: ghcr.io/linuxserver/pyload-ng:develop-71c2b445-ls19

Issue mentioned [here](https://github.com/linuxserver/docker-pyload-ng/issues/80#issuecomment-4243166907) but I was advised to post it here :)

GammaC0de:
FYI: dev97 was released few days ago

engels0n:
@GammaC0de 
yes, tried it but the bug still exists.
[...]

PR #4730, opened 2026-04-14
fix(ddownload): premium and valid_until pattern

dr3st:
### Describe the changes

Fixing the premium and premium valid until pattern.

see https://github.com/pyload/pyload/issues/4729


GammaC0de:

dr3st:
> Please increase plugin version

done, 0.10 now

PR #4731, opened 2026-04-15
POST request without valid auth does not return 401

Sab44:
### Describe the changes
When executing a POST API request with missing / invalid API key the response is
```
400
{"error": "CSRF token is invalid"}
```
when it should be 
```
401
{"error": "Invalid API credentials"}
```
for GET requests it is correct.

I have therefore moved the `csrf.protect()` so it is only executed if an authenticated session exists.
Otherwise we will always return `401`.

I don't think this causes any security concerns, but I'm no expert on security matters.

Incorrect API response, confusing for clients.

### DIFF
--- a/src/pyload/webui/app/helpers.py
+++ b/src/pyload/webui/app/helpers.py
@@ -387,9 +387,9 @@ def apikey_auth(func):
                 return flask.json.jsonify({"error": key_info["error"]}), 401
 
         # No API auth - still use the decorated function but rely on session auth
-        csrf.protect()
         s = flask.session
         if is_authenticated(s):
+            csrf.protect()
             user_info = {
                 "id": s["id"],
                 "name": s["name"],


Sab44:
Hi @GammaC0de 
thanks for addressing this so quickly.

Even with your changes though, a POST request that has a missing or empty API key header will not return a 401.
For example:
```
~ curl -X POST http://localhost:8000/api/restart -H 'X-API-Key: '
CSRF token is invalid
~ curl -X POST http://localhost:8000/api/restart
CSRF token is invalid
```
In my opinion this is quite confusing for anyone calling the API, as they would think they need to provide some kind of CSRF token in their request when in reality they just need to add an API key. So a `401` response code (with a suitable message) would be appropriate.

The underlying cause for this is that inside `csrf.protect()` there is this check:
```
        if request.method not in current_app.config["WTF_CSRF_METHODS"]:
            return
```
where `current_app.config["WTF_CSRF_METHODS"]` is `{'DELETE', 'POST', 'PUT', 'PATCH'}`. That is why it doesn't happen for GET requests, as they fulfill this condition and trigger an early return.

It's what I tried to address with this PR by moving the function call so it's only executed in case there is a valid session auth. But of course there might be a better / more secure solution.

GammaC0de:
`curl -v -X POST http://localhost:8000/api/restart -H 'X-API-Key: '` will not send empty  `X-API-Key` header:
```
* Host localhost:8000 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8000...
* Established connection to localhost (::1 port 8000) from ::1 port 10388
* using HTTP/1.x
> POST /api/restart HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/8.18.0
> Accept: */*
>
* Request completely sent off
```

Sab44:
> `curl -v -X POST http://localhost:8000/api/restart -H 'X-API-Key: '` will not send empty `X-API-Key` header:

Ah well, fair enough. What I was pointing at was that this check
```
api_key = flask.request.headers.get("X-API-Key")
if api_key:
```
would fail for any "falsy" value of `api_key` which could also be an empty string in addition to `None`. Maybe this can't actually happen, but it's better to have it covered.

Your newest commit fixed it now, thanks 🙂

PR #4734, opened 2026-04-18
allow drag-and-drop container upload to queue

fabianwimberger:
---
name: WebUI Update (pyLoad Next)
about: Propose an update for the web interface
title: ""
labels: pyLoad Next, enhancement
assignees: gammac0de
---

### Describe the changes

Drop a container file (`.dlc`, `.ccf`, `.rsdf`, `.txt`) anywhere on the web UI to add it to the queue directly, without opening the Add Package modal. The filename is used as the package name (backend already defaults to this when `add_name` is empty). Unsupported extensions are rejected client-side with a toast. Applies to both `modern` and `pyplex` themes.

Implementation: JS only, no backend or template changes. Uses the existing `/json/add_package` endpoint, relies on the global `$.ajaxSetup` for CSRF, and reuses the existing `indicateLoad`/`indicateSuccess`/`indicateFail` toast helpers. A transparent full-window overlay appears only while files are being dragged over the page.

Adding a container currently requires clicking **+**, filling in a package name (the form marks it `required`), picking a destination, and submitting — every time. One-step drag-and-drop is noticeably faster and sufficient for most usecases.

### Additional references

Builds on #4722, which added queue-destination support to `/json/add_package` via `upload_container`.


GammaC0de:
Awesome, thanks.

PR #4735, opened 2026-04-20
Fix tests and run in pipeline

Sab44:
### Describe the changes

Current unit tests had many issues. Many were broken altogether, for others there was no proper way of running them. I don't think they were useful in this state.

This PR fixes those tests:
- delete obsolete / completely broken tests
- fix tests that are still valid
- move unit tests in a `unit` subdirectory
- (re-)add test dependencies to `setup.cfg`
- refactor tests so all run via `pytest` (and not some pytest and some unittest)
- adapt dev setup instructions in README
- add test run instructions in README
- actually run the tests in the GitHub CI with coverage report
  - example run: https://github.com/Sab44/pyload/actions/runs/24653121323

Additionally, I added some integration tests in an `integration` subdirectory. These integration tests spin up a test pyLoad core instance and then verify API calls are processed properly. There are some basic tests for:
- REST API
- CNL API
- OpenAPI spec

I encourage extending the test suite over time to cover core functionality to ensure some robustness and avoid regressions / re-introducing bugs that were already fixed before.

Existing tests being pretty much useless.
And of course the problems that arise when there are no tests: refactoring becomes tricky and can introduce bugs, manual testing effort etc.

### Additional references

Integration tests use the test client provided by flask: https://flask.palletsprojects.com/en/stable/testing/


bin/system_check.py
```diff
@@ -93,18 +93,12 @@ def main():
             "Install OpenSSL if you want to create a secure connection to the core."
         )
 
-    if not js:
-        print("no JavaScript engine found")
-        print(
-            "You will need this for some Click'N'Load links. Install Spidermonkey, ossp-js, pyv8 or rhino"
-        )
-
```
Sab44, line 101: This script is not a unit test. I moved it into a `bin` subdirectory and removed this non-functional code.

src/pyload/core/utils/web/parse.py
```diff
@@ -118,6 +118,8 @@ def name(url, safe_name=True):
         name = us.query.split("=", 1)[::-1][0].split("&", 1)[0]
     if not name:
         name = ".".join(us.netloc.split(".")[:-1])
+    if us.fragment:
+        name += f"#{us.fragment}"
```
Sab44, line 122: This code is actually tested in the `test_parse.py` unit test. But since the test never ran it wasn't noticed that it failed. So I had to add this condition here to make the test pass again.

tests/unit/test_account_fixes.py
```diff
@@ -4,64 +4,7 @@
...
-# Now import data.py via importlib from the source tree
-_data_path = os.path.normpath(
-    os.path.join(os.path.dirname(__file__), "..", "src", "pyload", "core", "datatypes", "data.py")
-)
-_spec = importlib.util.spec_from_file_location("pyload.core.datatypes.data", _data_path)
-_data_mod = importlib.util.module_from_spec(_spec)
-sys.modules["pyload.core.datatypes.data"] = _data_mod
-_spec.loader.exec_module(_data_mod)
-
-AccountInfo = _data_mod.AccountInfo
```
Sab44, line 64: This is quite the vibe-coded mess. I fixed it now and the test runs properly, but I haven't checked whether these tests here are justified.

GammaC0de: Well, this test was actually introduced here https://github.com/pyload/pyload/pull/4706

GammaC0de:
Thanks again 👍 

EDIT: shouldn't the tests run on all python versions and platforms!?

tests/unit/test_account_fixes.py
Sab44: Oh yeah I know, just wanted to explain why I removed this entire block of code.

Sab44:
> EDIT: shouldn't the tests run on all python versions and platforms!?

I had this at first, there was a small issue on windows with permissions on the filesystem for the temporary files and it was also quite messy because every run would re-calculate the test coverage.

I'll see if I can refactor the workflow slightly so the tests run on every os / python version but the coverage is only calculated once.

PR #4740, opened 2026-04-26
Call secure_filename for folder when adding new package

Sab44:
### Describe the changes

The following problem happened to me:

1. Add a link with a package name that contains a space e.g. "My Test": `add_package` https://github.com/pyload/pyload/blob/d90a5f49802e2456a3e362cf73891ae978273399/src/pyload/core/api/__init__.py#L548 will not strip the spaces from the `folder` variable
2. Edit the package afterwards and e.g. enter a password for the package, then save it: `edit_package` https://github.com/pyload/pyload/blob/d90a5f49802e2456a3e362cf73891ae978273399/src/pyload/webui/app/blueprints/json_blueprint.py#L205 will overwrite the folder and strip any spaces from it: `pack_folder = secure_filename(pack_folder)`
3. If the file(s) already started to download, extraction will fail as it will try to find the file in `/downloads/My_Test/` when it is actually in `/downloads/My Test/`

This PR fixes this by calling `secure_filename` on the initial folder that is created when adding a package that will strip spaces (among others).

Added a test method for this as well to verify the behaviour.

I don't think this will have any side effects but I'm not 100% sure.

### DIFF
--- a/src/pyload/core/api/__init__.py
+++ b/src/pyload/core/api/__init__.py
@@ -559,17 +559,8 @@ class Api:
         else:
             folder = ""
 
-        folder = (
-            folder.replace("http://", "")
-            .replace("https://", "")
-            .replace("../", "_")
-            .replace("..\\", "_")
-            .replace(":", "")
-            .replace("/", "_")
-            .replace("\\", "_")
-            .replace("\r", "_")
-            .replace("\n", "_")
-        )
+        folder = folder.replace("http://", "").replace("https://", "")
+        folder = secure_filename(folder)
 
         sanitized_name = name.replace("\n", "\\n").replace("\r", "\\r")
         package_id = self.pyload.files.add_package(sanitized_name, folder, Destination(dest))
--- a/tests/integration/test_api.py
+++ b/tests/integration/test_api.py
@@ -170,3 +170,30 @@ class TestPackages:
         )
 
         assert response.status_code == 200
+
+    def test_package_folder_is_sanitized(self, api_key, client):
+        # Add a dummy package to the collector
+        response = client.post(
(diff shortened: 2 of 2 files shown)


Sab44:
I guess this is also a security concern, same as https://github.com/advisories/GHSA-6px9-j4qr-xfjw but here it is when adding a package, as the logic before this PR also only did a single-pass replacement on the folder path.

Not sure if it warrants a security advisory, never did one before.

EDIT: probably not the same issue here, as path elements like ../ are replaced with an underscore instead of an empty string.

GammaC0de:
> Not sure if it warrants a security advisory, never did one before.

GHSA-97r3-5w84-r4q8
I don't know what happened recently  but it seems like security researchers fell in love with pyLoad keeping me busy flooding with advisories

Sab44:
> [GHSA-97r3-5w84-r4q8](https://github.com/pyload/pyload/security/advisories/GHSA-97r3-5w84-r4q8) 
> I don't know what happened recently but it seems like security researchers fell in love with pyLoad keeping me busy flooding with advisories

Wow, no way that is a coincidence.

I assume since AI tools have become so good and widely available, people are constantly scanning repos and trying to find any "vulnerabilities" in order to be the first to disclose them. With these LLMs it is fairly easy to generate a report with small poc. I guess this is in order to pad the GitHub profiles for these "security researchers" (I'm sure there are some legit ones as well) with the amount of CVEs / vulnerabilities they have detected to get a better job / pay etc.

Anyway, for this particular one, if you read it closely it is actually quite sloppy and the proof of concept doesn't make sense. I guess I should've disclosed it myself lol, at least for some partial credit.

Issue #4762, opened 2026-06-22
Add CaptchaAI as a captcha-solver addon (sibling to AntiCaptcha)?


bshahin101:
Hi — pyLoad's addon design makes this clean. I'd like to contribute a CaptchaAI solver addon alongside AntiCaptcha.py, following the same plugin interface. CaptchaAI is an API-first solver (reCAPTCHA v2/v3 + Enterprise, Cloudflare Turnstile, GeeTest, image/OCR) with a 2Captcha-compatible endpoint, so it'd be a self-contained sibling plugin — same hook points, its own api_url/key config. Happy to send the PR matching your addon conventions + config schema if you'd welcome it. No pressure.

GammaC0de:
Fell Free to issue a PR. please note that the 2Captcha plugin was updated long time ago and may not be functional

PR #4764, opened 2026-06-23
[addons] Add CaptchaAI captcha solver addon

bshahin:
Adds a `CaptchaAI` addon that solves captchas via CaptchaAI's 2Captcha-compatible `in.php`/`res.php` API, addressing #4762 (the 2Captcha addon was noted as long-unmaintained / possibly non-functional).

Mirrors the structure and config conventions of the existing `AntiCaptcha` addon. Supported captcha types: **image/text (OCR)**, **reCAPTCHA v2 (normal + invisible)**, and **Cloudflare Turnstile** — matching the captcha types pyLoad's bundled anticaptcha plugins actually emit. hCaptcha and FunCaptcha are intentionally not included.

- Configurable `api_url` (default `https://ocr.captchaai.com`), `passkey`, per-type solve toggles, timeout, and optional bad-result reporting (`reportbad`).
- Submit -> poll (`action=get`) -> return token; balance check via `action=getbalance`.
- No new dependencies (uses the existing `BaseAddon` + `self.load`).

Status `testing`. Happy to adjust style/attribution to your preference.

### DIFF
--- a/src/pyload/plugins/addons/CaptchaAI.py
+++ b/src/pyload/plugins/addons/CaptchaAI.py
@@ -1,0 +1,242 @@
+import base64
+import time
+import urllib.parse
+
+from pyload.core.utils.convert import to_str
+
+from ..base.addon import BaseAddon, threaded
+
+
+class CaptchaAIException(Exception):
+    def __init__(self, err):
+        self.err = err
+
+    def get_code(self):
+        return self.err
+
+    def __str__(self):
+        return "<CaptchaAIException {}>".format(self.err)
+
+    def __repr__(self):
+        return "<CaptchaAIException {}>".format(self.err)
+
+
+class CaptchaAI(BaseAddon):
+    __name__ = "CaptchaAI"
+    __type__ = "addon"
+    __version__ = "0.01"
+    __status__ = "testing"
+
+    __config__ = [
+        ("enabled", "bool", "Activated", False),
+        ("check_client", "bool", "Don't use if client is connected", True),
+        ("solve_image", "bool", "Solve image captcha", True),
+        ("solve_recaptcha", "bool", "Solve ReCaptcha", True),
+        ("solve_turnstile", "bool", "Solve Turnstile", True),
+        ("refund", "bool", "Report bad result (refund) if result incorrect", False),
+        ("api_url", "str", "API base URL", "https://ocr.captchaai.com"),
+        ("passkey", "password", "API key", ""),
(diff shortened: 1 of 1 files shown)


bshahin:
Hi @GammaC0de — gentle check-in: the CLA is signed and the addon is ready whenever you have a moment. I think the CI checks are just waiting on a maintainer to approve the workflow run (first-time contributor), so if there's anything I can do to help it move, happy to. No rush — thanks again for the go-ahead!

GammaC0de:
Hi,
The file should reside in `src/pyload/plugins/anticaptchas/` and the class should be based on `CaptchaService`.

Would it be possible to get a user/API key for testing?

bshahin:
Thanks for taking a look, and good question on placement — let me clarify the intent, because I think the name may be a bit misleading.

CaptchaAI isn''t a captcha *type* — it''s a 2Captcha-compatible solving *service*, the same category as your own `AntiCaptcha` addon (and `DeathByCaptcha` / `Captcha9Kw`). Those all live in `plugins/addons/` as `BaseAddon` and register through `captcha_task()` / `task.handler.append(self)`. The `anticaptchas/` `CaptchaService` plugins (ReCaptcha, Turnstile, …) are the page-side detectors that hosters instantiate directly (e.g. `TurbobitNet` does `ReCaptcha(self.pyfile)`), so a solver backend placed there wouldn''t get constructed by the framework.

I modelled `addons/CaptchaAI.py` closely on your `AntiCaptcha` for exactly that reason. Did you want it moved regardless, or was the suggestion based on the name reading like a captcha type? Happy to restructure if you have a different design in mind — just want to avoid shipping a plugin the framework never instantiates.

On the test key: yes, I can get you one — I''ll sort out an API key for testing and follow up here.

GammaC0de:
Why HCaptcha not included?

bshahin:
Good question. CaptchaAI doesn''t currently offer hCaptcha solving — its supported set is reCAPTCHA v2/v3/Enterprise, Cloudflare Turnstile, and normal image captchas. I left hCaptcha out deliberately rather than register a handler that would always return an error for it.

Everything else is wired up the same way your `AntiCaptcha` addon is (reCAPTCHA + Turnstile + image, via `captcha_task()` / `task.handler.append`). If CaptchaAI adds hCaptcha support down the line, extending the plugin is trivial — just another handler branch — so this is easy to grow into later without blocking the rest now.

PR #4769, opened 2026-07-18
fix: race condition on shared account info causes KeyError during download

waldh4ri:
## Problem

`BaseAccount.choose()` clears `self.info` before resyncing it for the newly selected user:

```python
self.user = user
self.info.clear()
self.req.close()
...
if not self.logged:
    self.relogin()
```

The account plugin instance is shared across every download thread for a given hoster (`AccountManager.get_account_plugin` caches a single instance per plugin). Downloader plugins read `self.account.info["login"]["password"]` directly, without holding the account's lock.

When one thread runs `choose()` and clears `info` while another thread is mid-download reading from it, the reader can observe an empty dict and crash:

```
File "pyload/plugins/downloaders/AlldebridCom.py", line 83, in handle_premium
    "apikey": self.account.info["login"]["password"],
              ~~~~~~~~~~~~~~~~~^^^^^^^^^^
KeyError: 'login'
```

This affects every downloader that reads `account.info["login"][...]` directly (AlldebridCom, PremiumizeMe, RealDebrid, TorboxApp, FikperCom, ZeveraCom, and others), and shows up intermittently under concurrent downloads, causing some files in a batch to fail while others succeed.

A prior change (add some more locks) added `@lock` to `login()`, `logout()`, and `sync()`, but the readers in the downloader plugins never acquire that lock, so the empty-dict window was never actually closed.

## Fix

Replace `self.info.clear()` with `self.sync()`. `sync()` already runs a few lines later via the `self.logged` check, so the explicit clear is redundant. Calling `sync()` directly rebuilds `info["login"]`/`info["data"]` for the new user in place, so `self.info` transitions straight from the old user's data to the new user's data and is never left empty for a reader on another thread to observe.

### DIFF
--- a/src/pyload/plugins/base/account.py
+++ b/src/pyload/plugins/base/account.py
@@ -495,7 +495,7 @@ class BaseAccount(BasePlugin):
 
         else:
             self.user = user
-            self.info.clear()
+            self.sync()
             self.req.close()
 
             self.req = self.pyload.request_factory.get_request(


src/pyload/plugins/base/account.py
```diff
@@ -495,7 +495,7 @@ def choose(self, user=None):
 
         else:
             self.user = user
-            self.info.clear()
```
GammaC0de, line 498: i think calling clear() and then sync() would be better, please check and report

waldh4ri: That would reintroduce the race (with a shorter window) and could lead to the same `KeyError`. `self.info` update mechanism must be atomic because we don't acquire `self.account.lock` before reading `info` in any of `hoster.py`/`AlldebridCom.py`/etc. `sync` update to `info` is atomic and never leave it in a blank state:

```
d = {"login": {}, "data": {}}
for k, v in u.items():
    ...
self.info.update(d)
```

PR #4779, opened 2026-09-07
Add OrigridIo hoster plugin

mkiradani:
Adds a downloader plugin for **Origrid** (https://origrid.io), a large-file host.

**Full disclosure: I'm the operator of Origrid**, so please treat this as a vendor submission and verify independently — everything below can be checked with `curl` against a live link.

### Why it's short

No JS unpacking needed. Name, size and the direct link are all in the server-rendered HTML, and the free path has no captcha, no wait timer and no account requirement, so `SimpleDownloader` covers it with four patterns and a five-line `handle_free`.

### What it resolves, and how you can check it

Take any file page, e.g. `https://origrid.io/d/85k3lFZNL`:

- **Name** — `data-origrid="name"` on the `<h1>`
- **Size** — `data-origrid-bytes` carries the exact byte count, so there's no unit parsing or rounding
- **Offline** — HTTP 404 with `This link doesn't lead to a file`
- **Direct link** — `data-origrid="download"` on the anchor; it 302s to storage with `Accept-Ranges: bytes`, and range requests return `206 Partial Content`, so `resume_download` and chunking both work

### About the `data-origrid` attributes

They're a deliberate contract on the host's side, not incidental markup. Origrid's frontend is SvelteKit, whose CSS class names carry a build hash and change on every deploy — a plugin keyed on classes would break at the first release. The attributes are documented as public contract in their source and covered by a test on their side that fails if one is renamed or if the download anchor stops being server-rendered.

Mentioning it because it's the part most likely to rot silently, and it shouldn't.

### Notes

- `chunk_limit = -1` and `multi_dl = True`: the host caps neither concurrent downloads nor connections on the free path.
- No premium handling. Paid tiers exist but only change size limits and ads, not the download flow, so there's nothing for the plugin to do differently.

Happy to adjust anything to match your conventions.

### DIFF
--- a/src/pyload/plugins/downloaders/OrigridIo.py
+++ b/src/pyload/plugins/downloaders/OrigridIo.py
@@ -1,5 +1,3 @@
-# -*- coding: utf-8 -*-
-#
 # pyLoad hoster plugin for Origrid (https://origrid.io)
 #
 # Origrid serves everything a downloader needs from plain HTML: no captcha, no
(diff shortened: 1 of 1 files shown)


mkiradani:
Pushed a fix to this branch, found while testing against the live site rather than in review.

An **expired** file serves a different page from an unknown id, so it matched neither the online nor the offline patterns — pyLoad would have treated it as a temporary error and retried in a loop instead of marking the link dead. `OFFLINE_PATTERN` now covers both states.

It keys on a `data-origrid="offline"` attribute rather than the visible string, so translating or rewriting that copy can't silently break it. Same reasoning as the other anchors in this plugin.

Worth being upfront about why it slipped through: until today there were no expired files being served on our side at all, so that page state was unreachable and my tests couldn't have hit it. That changed this morning and the gap showed up immediately.

Also dropped the coding cookie (4 of 209 downloaders carry it) and added `URL_REPLACEMENTS` to normalise www/query-string variants, matching what the simpler plugins here do.

I'll get the CLA signed.

PR #4781, opened 2026-09-10
[DdownloadCom] Use HTTPS for plugin URL

oromeis:
DdownloadCom still uses `http://ddownload.com` as `PLUGIN_URL`.

DDownload redirects HTTP requests to HTTPS, but this causes the interactive
Turnstile captcha to fail when the pyLoad WebUI itself is accessed over HTTPS.

The browser blocks the initial `http://ddownload.com/` request as mixed active
content before the redirect to HTTPS can occur.

Changing `PLUGIN_URL` to `https://ddownload.com` fixes the interactive captcha
when pyLoad is accessed through an HTTPS reverse proxy.

Tested with pyLoad-ng running in Docker behind an HTTPS reverse proxy.


GammaC0de:

oromeis:
Plugin version has been bumped to 0.12. Thanks for the feedback!

